PR #34, opened 2022-07-08
Ftr: Migrate to protocol v2 for golang

pherzheyu:


aaron-ai (commented):
Maybe we should hide the generated code and provide doc for generating.

guyinyou (approved):

PR #107, opened 2022-07-31
Fix the naming issue

aaron-ai:


RockteMQ-AI (approved):
### Summary
Renames `ackFifoMessageLater` to `ackMessageLater` and updates log messages to say "message" instead of "fifo message", since the method handles both FIFO and non-FIFO acknowledgment paths.

**Changes reviewed:**
- **ProcessQueueImpl.java**: Method rename + log message updates — correct naming consistency fix. The method is called from both FIFO and standard message acknowledgment paths, so the generic name is more accurate.

**Note:** This PR has been open since July 2022. The change is trivial and safe, but please verify it still applies cleanly to the current `master` branch before merging.

**Verdict:** LGTM — simple naming consistency improvement.

---

PR #149, opened 2022-08-11
fix const SPAN_ATTRIBUTE_VALUE_MESSAGING_PROTOCOL_VERSION

guyinyou:
As far as I understand it should be the v2 protocol now


lizhanhui:
Nice catch. Thank you!

PR #155, opened 2022-08-16
fix bug in changeInvisibleDuration

guyinyou:
fix bug in changeInvisibleDuration


aaron-ai (commented):
Line is longer than 120 characters.

guyinyou:
> Nice catch! It would better if the checkstyle issue is addressed.

i will

Issue #179, opened 2022-08-25
Golang: Make each 'Client' have an independent 'ClientManager'


guyinyou:
Currently a process globally shares a 'ClientManager', so only one 'Conn' is established for each endpoint. When users create a new client, their expectations should be completely independent.

![image](https://user-images.githubusercontent.com/36399867/186558272-efd72b4a-3381-4e7c-a53f-ee8235cca14a.png)

leizhiyuan:
@guyinyou https://github.com/apache/rocketmq-clients/pull/235 could you reopen this pr？

guyinyou:
> @guyinyou #235 could you reopen this pr？

@aaron-ai

RockteMQ-AI:
**Issue Evaluation**

Category: `unclassified` | Status: **Needs Triage**

This issue could not be automatically classified. A maintainer will review it manually.

---
*Automated evaluation by RockteMQ-AI*

PR #235, opened 2022-09-12
feat: Make each 'Client' have an independent 'ClientManager'

leizhiyuan:
issue: #179

 Make each 'Client' have an independent 'ClientManager'


aaron-ai:
@guyinyou PTAL

leizhiyuan:
？

golang/client_manager.go
```diff
@@ -90,13 +90,14 @@ type ClientManagerRegistry interface {
 }
 
 type clientManagerRegistry struct {
```
aaron-ai, line 92: Could we remove `clientManagerRegistry`, since each `client` has their own `clientManager`, it seems no need to maintain a "registry".

aaron-ai:
@leizhiyuan Thanks for your attention! Sorry for not responding to this PR for a long time. 

I am not familiar with Golang. IMHO, maybe we could construct a `ClientManager` directory if the singleton pattern is not applied in `ClientManager`, so the "registry" is seemingly redundant.

guyinyou (commented):
lgtm

guyinyou (approved):

aaron-ai:
Thanks @leizhiyuan

PR #264, opened 2022-10-24
Add extraProperties for PublishingMessageImpl

aaron-ai:
Since OpenTelemetry community recommended to propagate trace context through message properties rather than the existing `trace_context` field, and the implementation of `PublishingMessageImpl#getProperties` made a deep copy, making it impossible to update the message properties in `PublishingMessageImpl`, an independent method should be added for propagating trace context through `PublishingMessageImpl`.


aaron-ai:
related issue: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/6884

zhouxinyu (approved):

PR #272, opened 2022-11-01
Define index of LoadBalancer as static

aaron-ai:
There is no need to generate index in constructor of load balancer.


lollipopjin (approved):
LGTM

Issue #255, opened 2022-09-29
Implement native php client


zhouxinyu:

security00:
Please assign to me, thanks

security00:
Please assign to me, thanks

zhouxinyu:
> Please assign to me, thanks

Done, thanks for your interest. We already have some [docs](https://github.com/apache/rocketmq-clients/tree/master/docs) to help you implement the client.

Any questions please feel free to start a discussion on `Issues` or mailing lists.

lizhanhui:
Hi @security00 , Great to see that you have interests... Feel free to get feedback.

security00:
> Hi @security00 , Great to see that you have interests... Feel free to get feedback.

ok, I need a rocketMQ5 test server now, can you help me to provide it, thanks.

PR #278, opened 2022-11-09
Implement native php client #255

security00:
Implement native php client #255


php/Producer.php
```diff
@@ -0,0 +1,70 @@
...
+use Apache\Rocketmq\V2\Resource;
+use Grpc\ChannelCredentials;
+use const Grpc\STATUS_OK;
+
+class Producer
+{
+
+    public function init()
+    {
+        // 客户端ID 因为目前我的主机名称是中文的，害怕解析会有问题，所以暂时先将主机名称写死
```
lizhanhui, line 37: We are serving users across the globe...Best to comment in English

php/composer.lock
```diff
@@ -0,0 +1,1470 @@
+{
```
lizhanhui, line 1: Do we need to submit this file to the repository? I guess it is generated by IDE according to file `composer.json` to guard against concurrent writes

lizhanhui:
Let's separate sources automatically generated by `protoc` from these manually maintained.

php/Producer.php
security00: Okay, I'll use English comments instead of the Chinese comments that appear in the code

php/composer.lock
security00: Okay, I'll remove this file from the repository

security00:
> Let's separate sources automatically generated by `protoc` from these manually maintained. It looks folder `php/grpc/Aapche/RocketMQ/V2` has both types of source files.

ok

php/composer.lock
aaron-ai: I create a pull request to add all related files into .gitignore, see #280

php/composer.json
```diff
@@ -0,0 +1,13 @@
+{
+  "require":{
+    "google/protobuf": "^3.3",
+    "grpc/grpc": "^1.42",
+    "hanson/foundation-sdk": "^4.0"
```
aaron-ai, line 5: It seems that foundation-sdk is not compatiable with PHP 8+, upgrade it to 5.x?

lizhanhui:
I am fine with the current status quo... +1 to merge it as a starting point

php/composer.json
security00: This component is used to assist in the development of the SDK, it is not used yet, if you do not need it later, you can directly remove it from the composer.json

php/Producer.php
bendan365: 'update_metadata' => function ($metaData) use ($clientId) {
            $metaData['headers'] = ['clientID' => $clientId]; // Pass the ClientID to the server through the header 
            return $metaData;
        }
but the response data is 
{"message":"\n\u001f\b\u0478\u0002\u0012\u0019client id cannot be empty","status":{"metadata":[],"code":0,"details":""},"metadata":[]}


$metaData['headers'] = ['x-mq-client-id' => $clientId];
the response data is the same.

PR #324, opened 2022-12-23
dotnet driver: Fix message queue handling

colprog:


csharp/rocketmq-client-csharp/SimpleConsumer.cs
```diff
@@ -196,12 +196,13 @@ internal override void OnSettingsReceived(rmq.Settings settings)
...
 
-        public async Task<List<Message>> Receive(int batchSize, TimeSpan invisibleDuration)
+        public async Task<List<Message>> Receive(int batchSize, TimeSpan invisibleDuration, TimeSpan? awaitDuration = null)
         {
             var messageQueue = NextQueue();
             if (null == messageQueue)
             {
-                Logger.Debug("NextQueue returned null");
+                Logger.Debug("NextQueue returned null, waiting for 50ms");
+                await Task.Delay(50);
```
aaron-ai, line 205: Why did we need to delay here?

csharp/rocketmq-client-csharp/Producer.cs
```diff
@@ -128,22 +128,18 @@ public async Task<SendReceipt> Send(Message message)
...
                     ResourceNamespace = resourceNamespace(),
                     Name = message.Topic
-                }
+                },
+                UserProperties = { message.UserProperties },
+                SystemProperties = new rmq::SystemProperties
+                {
+                    MessageId = message.MessageId,
+                    MessageType = rmq::MessageType.Normal,
+                    Tag = message.Tag ?? string.Empty,
```
aaron-ai, line 137: If message's tag is not set, we should not set it as an empty string.

As you could see from https://github.com/apache/rocketmq-apis/blob/e83a1bceacbef5be465a459965d8edee59cad0c2/apache/rocketmq/v2/definition.proto#L200, the message tag field is optional, which means that it should be absent if message tag is not set.

aaron-ai:
Overall this PR looks solid, I left a couple of comments above, thanks @colprog

csharp/rocketmq-client-csharp/SimpleConsumer.cs
colprog: If we start the consumer without any topics. a receive loop became a busy loop with no yield point. this will saturate a client thread and takes up server resources.

csharp/rocketmq-client-csharp/Producer.cs
colprog: fixed in latest revision

csharp/rocketmq-client-csharp/SimpleConsumer.cs
aaron-ai: @colprog If the queue could not be chosen, an exception is preferred rather than an empty message list. Exception make developers could be aware of the unexcepted occation and it is within their discretion. On the other hand, consumer need to do some precheck before startup, including the subscription is set or not.

colprog: Making Start() lazy is probably out of scope for this changes.
Is throwing an exception here ok with you?

aaron-ai: > Making Start() lazy is probably out of scope for this changes. Is throwing an exception here ok with you?

Yes, I agree with it!

colprog: fixed in latest revision

PR #335, opened 2022-12-30
Doc: Rust README.md and Cargo.toml

CherishCai:
- remove main.rs
- add README.md
- add features in Cargo.toml


CherishCai:
如果要用新版 tonic 和 prost ，估计还是 test build proto 更佳了，先回滚。

rust/README.md
```diff
@@ -0,0 +1,43 @@
+# The Rust Implementation of Apache RocketMQ Client
+
+Here is the rust implementation of the client for [Apache RocketMQ](https://rocketmq.apache.org/).
+It is still at a very early stage of development.
+
+## Prerequisites
+
+| Stage   | Requirements |
+| ------- |--------------|
+| Build   | Rust 1.66+   |
```
aaron-ai, line 10: Could you clarify why we support this version rather than the previous one?

CherishCai: It simply defines the latest stable version at the current.

Of course, the previous version can be supported, but since we will gradually improve it from now on, let's use the latest version.

aaron-ai: @CherishCai Generally, the minimal version we support is largely consistent with the version that gRPC supports.

I am not a rust expert. Is it possible that some developers think we only support 1.66+ from the readme? If so, the compatibility of the RocketMQ rust client may be difficult to keep after a library that requires 1.66+ is introduced by someone and widely used.

CherishCai: The new version of the grammar candy is very good, and there is not much use of advanced features. At that time, if there is a demand for expanding the scope of use, simple modifications should be available.

As far as I know, Rust is simple and common. In fact, it is not long ago that many library support 1.49+ 1.56+.

So it's not a big problem. Improve it first and then consider the version support according to the user's issue

CherishCai: @aaron-ai I am also not a rust expert.
Not long ago, I contributed `nacos-sdk-rust` to get started. It should be OK for users to use it. It is easy to upgrade the new version of Rust.

CherishCai: @aaron-ai What changes need to be made? If possible, I will refer to Java to define the following apis.

lizhanhui: @CherishCai We should use the edition here as we do not have a specific requirement of the rust version.

CherishCai: > We should use the edition here as we do not have a specific requirement of the rust version.

emmm. How should it be modified? can you give some advice?

Is there a DingTalk group? Some of our friends can participate in the research and development, and it can also be used as a hands-on practice of Rust.

rust/src/lib.rs
```diff
@@ -14,13 +14,17 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#[path = "pb/apache.rocketmq.v2.rs"]
```
lizhanhui, line 17: @CherishCai Why include! over specifying path?
rust-analyzer works better, say jump to definition if specifying mod by path.

rust/README.md
CherishCai: delete Prerequisites

PR #342, opened 2023-01-16
Add API of PullConsumer

### LINKED ISSUE #341: Add API of PullConsumer

aaron-ai:
Similar to the `LitePullConsumer` provided in RocketMQ 4.0, RocketMQ 5.0 will also provide a brand-new API that can achieve the same functionality, providing higher controllability and flexibility to meet more diverse requirements for message processing.

Fixes #341

### DIFF
--- a/java/client-apis/src/main/java/org/apache/rocketmq/client/apis/consumer/PullConsumer.java
+++ b/java/client-apis/src/main/java/org/apache/rocketmq/client/apis/consumer/PullConsumer.java
@@ -57,6 +57,8 @@ public interface PullConsumer extends Closeable {
     /**
      * Fetch messages from assigned message queues specified by {@link #assign(Collection)}.
      *
+     * <p>The messages polled from remote are across the message queue.
+     *
      * @param timeout the maximum time to block.
      * @return list of fetched messages.
      */
--- a/java/client-apis/src/main/java/org/apache/rocketmq/client/apis/consumer/PullConsumer.java
+++ b/java/client-apis/src/main/java/org/apache/rocketmq/client/apis/consumer/PullConsumer.java
@@ -1,0 +1,134 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.rocketmq.client.apis.consumer;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.time.Duration;
+import java.util.Collection;
+import java.util.List;
+import java.util.Optional;
+import org.apache.rocketmq.client.apis.ClientException;
+import org.apache.rocketmq.client.apis.message.MessageQueue;
+import org.apache.rocketmq.client.apis.message.MessageView;
+
+public interface PullConsumer extends Closeable {
+    /**
+     * Get the consumer group of the consumer.
+     */
+    String getConsumerGroup();
+
+    /**
+     * @param topic    the topic that needs to be monitored.
+     * @param listener the callback to detect the message queue changes.
+     */
+    void registerMessageQueueChangeListenerByTopic(String topic, TopicMessageQueueChangeListener listener);
+
+    /**
+     * Fetch message queues of the topic.
+     */
+    Collection<MessageQueue> fetchMessageQueues(String topic) throws ClientException;
+
+    /**
+     * Manually assign a list of message queues to this consumer.
+     *
+     * <p>This interface does not allow for incremental assignment and will replace the previous assignment (if
+     * previous assignment existed).
+     *
+     * @param messageQueues the list of message queues that are to be assigned to this consumer.
+     */
+    void assign(Collection<MessageQueue> messageQueues);
+
+    /**
+     * Fetch messages from assigned message queues specified by {@link #assign(Collection)}.
+     *
+     * @param timeout the maximum time to block.
+     * @return list of fetched messages.
+     */
+    List<MessageView> poll(Duration timeout);
+
+    /**
+     * Overrides the fetch offsets that the consumer will use on the next poll. If this method is invoked for the same
+     * message queue more than once, the latest offset will be used on the next {@link #poll(Duration)}.
+     *
+     * @param messageQueue the message queue to override the fetch offset.
+     * @param offset       message offset.
+     */
+    void seek(MessageQueue messageQueue, long offset);
+
+    /**
+     * Suspending message pulling from the message queues.
+     *
+     * @param messageQueues message queues that need to be suspended.
+     */
+    void pause(Collection<MessageQueue> messageQueues);
+
+    /**
+     * Resuming message pulling from the message queues.
+     *
+     * @param messageQueues message queues that need to be resumed.
+     */
+    void resume(Collection<MessageQueue> messageQueues);
+
+    /**
+     * Look up the offsets for the given message queue by timestamp. The returned offset for each message queue is the
+     * earliest offset whose timestamp is greater than or equal to the given timestamp in the corresponding message
+     * queue.
+     *
+     * @param messageQueue message queue that needs to be looked up.
+     * @param timestamp    the timestamp for which to search.
+     * @return the offset of the message queue, or {@link Optional#empty()} if there is no message.
+     */
+    Optional<Long> offsetForTimestamp(MessageQueue messageQueue, Long timestamp);
+
+    /**
+     * Get the latest committed offset for the given message queue.
+     *
+     * @return the latest committed offset, or {@link Optional#empty()} if there was no prior commit.
+     */
+    Optional<Long> committed(MessageQueue messageQueue);
+
+    /**
+     * Commit offset manually.
+     */
+    void commit() throws ClientException;
+
+    /**
+     * Overrides the fetch offsets with the beginning offset that the consumer will use on the next poll. If this
+     * method is invoked for the same message queue more than once, the latest offset will be used on the next
+     * {@link #poll(Duration)}.
+     *
+     * @param messageQueue the message queue to seek.
+     */
+    void seekToBegin(MessageQueue messageQueue) throws ClientException;
+
+    /**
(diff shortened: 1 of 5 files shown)


java/client-apis/src/main/java/org/apache/rocketmq/client/apis/message/MessageQueue.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.rocketmq.client.apis.message;
+
+public interface MessageQueue {
+    String getTopic();
+
+    String getId();
```
lizhimins, line 23: how about add Perm getPerm();

zhouxinyu: It's not a good idea since we shouldn't propagate the `Perm` concept here from the Ops level.

java/client-apis/src/main/java/org/apache/rocketmq/client/apis/consumer/PullConsumer.java
```diff
@@ -0,0 +1,134 @@
...
+     *
+     * @param messageQueues the list of message queues that are to be assigned to this consumer.
+     */
+    void assign(Collection<MessageQueue> messageQueues);
+
+    /**
+     * Fetch messages from assigned message queues specified by {@link #assign(Collection)}.
+     *
+     * @param timeout the maximum time to block.
+     * @return list of fetched messages.
```
zhouxinyu, line 61: Could you please clarify whether the returned messages are across the message queue or topic?

aaron-ai: Yes, it works across message queues/topics, and unless otherwise specified, the API is aimed at collections rather than individual queues/topics.

socutes:
In the scenario of stream processing, there are several methods to obtain offset information that are often used, and it is recommended to add them:
1.Gets the current consumption progress point. This is often used when the application crashes, data checks, and when obtaining the next consumption point.  eg: position()
2.Gets the earliest offset of the topic. eg: beginningOffsets()
3.Gets the offset at the end of the topic. eg: endOffsets()
@zhouxinyu @aaron-ai

java/client-apis/src/main/java/org/apache/rocketmq/client/apis/consumer/PullConsumer.java
zhouxinyu: I mean clarify it in the API doc 😄

aaron-ai: OK

aaron-ai:
> In the scenario of stream processing, there are several methods to obtain offset information that are often used, and it is recommended to add them: 1.Gets the current consumption progress point. This is often used when the application crashes, data checks, and when obtaining the next consumption point. eg: position() 2.Gets the earliest offset of the topic. eg: beginningOffsets() 3.Gets the offset at the end of the topic. eg: endOffsets() @zhouxinyu @aaron-ai

Make sense. At the same time, would it be more appropriate to place these methods in the Ops related interface rather than the pull consumer?

zhouxinyu:
@socutes I noticed that we already have some methods support seek to begin or end of a specifc message queue. These methods provide `query and seek` semantic, do we really need the query-only methods?

java/client-apis/src/main/java/org/apache/rocketmq/client/apis/consumer/PullConsumer.java
```diff
@@ -0,0 +1,136 @@
...
+     * Get the latest committed offset for the given message queue.
+     *
+     * @return the latest committed offset, or {@link Optional#empty()} if there was no prior commit.
+     */
+    Optional<Long> committed(MessageQueue messageQueue);
+
+    /**
+     * Commit offset manually.
+     */
+    void commit() throws ClientException;
```
lizhimins, line 111: commit 的语义不明确，比如说对于单个队列客户端 pull 进度是 200，poll 此时返回了 100，我们是不知道消费进度的，此时 commit 200 那就丢消息了，commit 100 也不对，对用户来说并不是说前一批消息消费完才能来拿下一批。所以这里的含义是 auto commit。对于流框架来说，通常希望 commit 和 ckpt 是原子的，我认为还应该增加 commit(mq, offset) 作为手动接口。

lizhimins: The semantics of commit are unclear, for example, for a single queue client, the pull progress is 200, and the poll function returns 100 at this time, we don't know the consumption progress. If we commit 200 will lose the message, commit 100 is not correct either. For users, it is not the case that the previous batch of messages needs to be consumed before taking the next batch. So the meaning here is auto commit. For stream frameworks, it is usually expected that commit and ckpt are atomic, and I think commit(mq, offset) should also be added as a manual interface.

java/client-apis/src/main/java/org/apache/rocketmq/client/apis/consumer/PullConsumer.java
```diff
@@ -0,0 +1,136 @@
...
+
+    /**
+     * Fetch messages from assigned message queues specified by {@link #assign(Collection)}.
+     *
+     * <p>The messages polled from remote are across the message queue.
+     *
+     * @param timeout the maximum time to block.
+     * @return list of fetched messages.
+     */
+    List<MessageView> poll(Duration timeout);
```
aaron-ai, line 65: Should we also provide an asynchronous interface for PullConsumer#poll? If we only offer a synchronous interface, it means that each call will occupy one user thread. Typically, the poll interface is called for multiple queues, so occupying one thread may not have a significant impact. However, an asynchronous interface can provide greater flexibility, such as forming a future chain with the user's own methods, reducing unnecessary threads when multiple clients exist in one process, and potentially offering other benefits.

We welcome everyone to join the discussion.

oss-sentinel-ai (approved):
## Review: Approved ✅

**PR:** #342 — Add API of PullConsumer
**Type:** API addition (3 files, +282/-2)

### Assessment
This PR introduces the PullConsumer API for RocketMQ 5.0, similar to `LitePullConsumer` in RocketMQ 4.0. It provides higher controllability and flexibility for diverse message processing requirements.

### Key Changes
- New `PullConsumer` interface (136 lines)
- New `PullConsumerBuilder` interface (82 lines)
- `TopicMessageQueueChangeListener` for queue change events (32 lines)
- `MessageQueue` abstraction (30 lines)

### Verdict
✅ Clean API design that follows existing patterns. Fixes #341.

---

RockteMQ-AI:

### Summary
Adds PullConsumer API interfaces for RocketMQ 5.0, providing manual message queue assignment and pull-based consumption similar to LitePullConsumer from 4.0, along with supporting builder, listener, and MessageQueue interfaces.

### Findings
- **[Critical]** PullConsumer.java:44-128 — Multiple Javadoc `@param` tags are malformed, using file paths like `@repos/apache_rocketmq-clients/cpp/...` instead of parameter names. This will break Javadoc generation and make the API documentation unusable.
- **[Critical]** PullConsumerBuilder.java:34-75 — All `@param` tags contain malformed file path references instead of parameter names, preventing proper Javadoc generation.
- **[Critical]** TopicMessageQueueChangeListener.java:28 — Malformed `@param` tags with file path references instead of parameter names.
- **[Warning]** PullConsumer.java:68,76,119,128 — Javadoc `{@link}` syntax is broken, referencing external files instead of using standard Java link syntax like `{@link #poll(Duration)}`.
- **[Warning]** PullConsumer.java:103 — Javadoc references `Optional#empty()` with broken syntax; should be `{@link Optional#empty()}`.
- **[Warning]** PullConsumer.java:96 — `offsetForTimestamp(MessageQueue messageQueue, Long timestamp)` uses boxed `Long` instead of primitive `long`. If null is not a valid input, use primitive for clarity and performance.
- **[Info]** PullConsumer.java:56,84 — `assign()`, `pause()`, and `resume()` methods don't document behavior for null or empty collections, or for queues not currently assigned.
- **[Info]** PullConsumer.java — No thread-safety documentation. Specify whether methods can be called concurrently from multiple threads.
- **[Info]** PullConsumer.java:68 — `poll(Duration timeout)` doesn't declare checked exceptions. Document what happens on network errors, consumer closure, or interruption (does it throw, return empty list, etc?).
- **[Info]** MessageQueue.java — Interface will be used in Collections (see `Set<MessageQueue>` in listener). Document `equals()` and `hashCode()` contract expectations.
- **[Info]** PullConsumer.java — No method to query current assignment or check if a specific queue is paused. Consider if these are needed for debugging/monitoring.

### Suggestions
1. **Fix all Javadoc tags immediately** — Replace all malformed `@repos/...` and `@gh_2.65.0_linux_amd64/...` references with proper parameter names and `{@link}` syntax. This is a blocker for the PR.

2. **Add null-safety documentation** — Specify behavior when null or empty collections are passed to `assign()`, `pause()`, `resume()`.

3. **Document thread safety** — Add class-level Javadoc stating whether the interface is thread-safe and which methods can be called concurrently.

4. **Clarify exception behavior** — Document what exceptions `poll()` might throw or return on error conditions.

5. **Consider adding query methods** — Methods like `assignment()` to get current assignments or `isPaused(MessageQueue)` could be valuable for debugging and monitoring, though not critical for initial implementation.

6. **Use primitive types where appropriate** — Change `Long timestamp` to `long timestamp` in `offsetForTimestamp()` unless null is a valid input that needs special handling.

---

RockteMQ-AI (commented):
### Summary
This PR modifies 6 file(s) with 330 lines of diff. No test changes detected — consider adding test coverage.

---

java/client-apis/src/main/java/org/apache/rocketmq/client/apis/consumer/PullConsumer.java
```diff
@@ -0,0 +1,136 @@
+/*
```
RockteMQ-AI, line 1: No test changes detected alongside source modifications. Consider adding tests to cover the changes.

RockteMQ-AI (approved):
﻿### Summary
This PR introduces the PullConsumer API for RocketMQ 5.0 Java client, analogous to Kafka's consumer API. The design includes:

- **PullConsumer**: Core interface with assign/poll/seek/pause/resume/commit operations
- **PullConsumerBuilder**: Builder pattern consistent with existing RocketMQ 5.0 client APIs
- **TopicMessageQueueChangeListener**: Callback for queue topology changes
- **MessageQueue**: Simple topic + id abstraction

The API design is clean, well-documented, and follows established messaging client patterns. This is an API-only PR (no implementation), which is appropriate for establishing the contract before implementation work.

Minor observations:
- `setMaxCacheMessageCountEachQueue` / `setMaxCacheMessageSizeInBytesEachQueue` — the "EachQueue" suffix is slightly unusual; consider whether these should be per-consumer or per-queue limits
- `seekToBegin` / `seekToEnd` throw `ClientException` while `seek(long offset)` does not — consider consistency

LGTM as an API proposal. Note this PR has been open since January 2023 — if the implementation is being actively worked on, please update the status.

---
*Automated review by "github-manager-bot"*

PR #350, opened 2023-02-05
add a basic producer demo

leizhiyuan:
add a basic implement for producer with rust

there are many code which need to improve, please give more review advice. 

image

with setting the value in proxy

    private boolean tlsTestModeEnable = false;
    public static TlsMode tlsMode = TlsMode.parse(System.getProperty(TLS_SERVER_MODE, "disabled"));


leizhiyuan:
now ,rust can not support tls ,there are some problems I do not resolve.

rust/src/client.rs
```diff
@@ -91,6 +111,117 @@ struct SessionManager {
...
+        srequest.messages.push(pb::Message {
+            topic: Some(Resource {
+                resource_namespace: "".to_string(),
+                name: message.topic.clone(),
+            }),
+            user_properties: Default::default(),
+            system_properties: Option::from(pb::SystemProperties {
+                tag: Option::from(message.tags.clone()),
+                keys: message.keys.clone(),
+                message_id: "123".to_string(),
```
aaron-ai, line 150: It would be great if the message id could be generated according to the rules [here](https://github.com/apache/rocketmq-clients/blob/master/docs/message_id.md).

aaron-ai:
I'm sorry, I just started reviewing this PR because I'm not very familiar with Rust. It would be even better if other peers who are familiar with Rust could participate as well.

rust/src/client.rs
leizhiyuan: > It would be great if the message id could be generated according to the rules [here](https://github.com/apache/rocketmq-clients/blob/master/docs/message_id.md).

done，someone can run the test case to verify it

rust/src/client.rs
```diff
@@ -89,6 +111,119 @@ impl Session {
 struct SessionManager {
     logger: Logger,
     tx: tokio::sync::mpsc::Sender<command::Command>,
+    selector: QueueSelector,
```
lizhanhui, line 114: `QueueSelector` makes sense to Producer only. Best to move to the Producer struct.

rust/src/client.rs
```diff
@@ -89,6 +111,119 @@ impl Session {
...
+}
+
+impl SessionManager {
+    async fn send(
+        &self,
+        message: &models::message::MessageImpl,
+        client: Weak<&Client>,
+    ) -> Result<SendMessageResponse, ClientError> {
```
lizhanhui, line 122: SessionManager helps manage sessions: by maintaining heartbeat, and administrative controlling flow...

Specific send steps should go to `Session` and `SessionManager` should directly forward workload to a target session after looking up the route-table: mapping between topic/queue to host addresses.

PR #358, opened 2023-02-10
Fix send message error when MessageType unspecified.

echooymxq:
Even if the proxy config set  `enableTopicMessageTypeCheck` false, sending a message always throw an error.

### DIFF
--- a/java/client/src/main/java/org/apache/rocketmq/client/java/message/MessageType.java
+++ b/java/client/src/main/java/org/apache/rocketmq/client/java/message/MessageType.java
@@ -21,7 +21,8 @@ public enum MessageType {
     NORMAL,
     FIFO,
     DELAY,
-    TRANSACTION;
+    TRANSACTION,
+    MESSAGE_TYPE_UNSPECIFIED;
 
     public static MessageType fromProtobuf(apache.rocketmq.v2.MessageType messageType) {
         switch (messageType) {
@@ -33,9 +34,8 @@ public enum MessageType {
                 return MessageType.DELAY;
             case TRANSACTION:
                 return MessageType.TRANSACTION;
-            case MESSAGE_TYPE_UNSPECIFIED:
             default:
-                throw new IllegalArgumentException("Message type is not specified");
+                return MessageType.MESSAGE_TYPE_UNSPECIFIED;
         }
     }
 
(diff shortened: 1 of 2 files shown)


aaron-ai:
@echooymxq  Producer will get the `settings` from the proxy, which contains the config of `validateMessageType`. you can debug the settings the client received.

echooymxq:
> 



> @echooymxq Producer will get the `settings` from the proxy, which contains the config of `validateMessageType`. you can debug the settings the client received.

@aaron-ai Hi, what i mean is if the proxy return the topic unspecified  message type(in most cases, never will be happen), even though not validate message type, it always send message error. you can also see the golang implementation.

aaron-ai:
@echooymxq Actually route data of each topic should have its message type explicitly, could you describe it clearly?

When does the proxy return unspecified message type?

echooymxq:
@aaron-ai thank your patient reply, there are some potential point ???
```java
// ClusterMetadataService#getTopicMessageType
public TopicMessageType getTopicMessageType(String topic) {
      TopicConfigAndQueueMapping topicConfigAndQueueMapping;
      try {
          topicConfigAndQueueMapping = topicConfigCache.get(topic);
      } catch (Exception e) {
         // Potential point ???
          return TopicMessageType.UNSPECIFIED;
      }
      if (topicConfigAndQueueMapping.equals(EMPTY_TOPIC_CONFIG)) {
          // Potential point ???
          return TopicMessageType.UNSPECIFIED;
      }
      return topicConfigAndQueueMapping.getTopicMessageType();
  }
```

echooymxq:
i will close this pr firstly until you think it's necessary.

aaron-ai:
@echooymxq you could create an issue to track it in [rocketmq's repo](https://github.com/apache/rocketmq). In rocketmq 5.0, route data of each topic should have its message type explicitly, so `UNSPECIFIED` should never be considered.

drpmma:
> @aaron-ai thank your patient reply, there are some potential point ???
> 
> ```java
> // ClusterMetadataService#getTopicMessageType
> public TopicMessageType getTopicMessageType(String topic) {
>       TopicConfigAndQueueMapping topicConfigAndQueueMapping;
>       try {
>           topicConfigAndQueueMapping = topicConfigCache.get(topic);
>       } catch (Exception e) {
>          // Potential point ???
>           return TopicMessageType.UNSPECIFIED;
>       }
>       if (topicConfigAndQueueMapping.equals(EMPTY_TOPIC_CONFIG)) {
>           // Potential point ???
>           return TopicMessageType.UNSPECIFIED;
>       }
>       return topicConfigAndQueueMapping.getTopicMessageType();
>   }
> ```
> 
> even if return the unspecified MessageType, if we set the validateMessageType false, send message maybe success.

The code means it is not TopicConfig in the cache. The getTopicMessageType will return NORMAL type if the type is unset.

Issue #182, opened 2022-08-25
Golang: Try to recover the When the observer is release


guyinyou:
When the observer is shut down, try to recover (after the next routing information update, check if the endpoints still exist, and then call the telemeter function)

![image](https://user-images.githubusercontent.com/36399867/186559696-68d20a23-9290-4e98-a1c8-5cfcd54e0add.png)

Anakin100100:
@guyinyou can we close this issue?

PR #408, opened 2023-03-17
try to recover on server TelemetryCommand transmission error

Anakin100100:
I'm trying to solve the issue described by @guyinyou in #182 

The new flow of execution after transmission error looks as follows:
 - we encounter a server TelemetryCommand transmission error
 - we set the state of recovery to true and skip to the next loop iteration
 - if there is a transmission error when we are in recovery we release the observer and set the state of recovery to false
 - if we manage to read the message we check if each of the servers is healthy, if not we release the observer and set the state of recovery to false
 - if we managed to read the message and each server is healthy we process the message and set the state of recovery to false. 

I'm going to write the unit tests for this next.

### DIFF
--- a/golang/.gitignore
+++ b/golang/.gitignore
@@ -1,0 +1,2 @@
+# Go test binaries
+*.test
--- a/golang/.gitignore
+++ b/golang/.gitignore
@@ -1,2 +1,0 @@
-# Go test binaries
-*.test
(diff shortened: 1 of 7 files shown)


guyinyou (approved):
LGTM

golang/.gitignore
```diff
@@ -0,0 +1,2 @@
+# Go test binaries
+*.test
```
aaron-ai, line 2: Let's move it to the top level.

Anakin100100: applied

aaron-ai:
Thanks @Anakin100100 !

Issue #294, opened 2022-11-30
golang DATA RACE


tianping526:
```golang
// github.com/apache/rocketmq-clients/golang@v0.0.0-20221125070211-5ba0ba13e773/rpc_client.go:98
// github.com/apache/rocketmq-clients/golang@v0.0.0-20221125070211-5ba0ba13e773/rpc_client.go:98
func (rc *rpcClient) QueryRoute(ctx context.Context, request *v2.QueryRouteRequest) (*v2.QueryRouteResponse, error) {
	rc.activityNanoTime = time.Now() // DATA RACE
	resp, err := rc.msc.QueryRoute(ctx, request)
	sugarBaseLogger.Debugf("queryRoute request: %v, response: %v, err: %v", request, resp, err)
	return resp, err
}

==================
==================
// github.com/apache/rocketmq-clients/golang@v0.0.0-20221125070211-5ba0ba13e773/rpc_client.go:137
// github.com/apache/rocketmq-clients/golang@v0.0.0-20221125070211-5ba0ba13e773/rpc_client.go:137
func (rc *rpcClient) ReceiveMessage(ctx context.Context, request *v2.ReceiveMessageRequest) (v2.MessagingService_ReceiveMessageClient, error) {
	rc.activityNanoTime = time.Now() // DATA RACE
	resp, err := rc.msc.ReceiveMessage(ctx, request)
	sugarBaseLogger.Debugf("receiveMessage request: %v, err: %v", request, err)
	return resp, err
}

==================
==================

// github.com/apache/rocketmq-clients/golang@v0.0.0-20221125070211-5ba0ba13e773/metric.go:166
func (dcmp *defaultClientMeterProvider) isEnabled() bool {
	return dcmp.clientMeter.enabled // DATA RACE
}

// github.com/apache/rocketmq-clients/golang@v0.0.0-20221125070211-5ba0ba13e773/metric.go:183
	if !metric.GetOn() {
		dcmp.clientMeter.shutdown()
		sugarBaseLogger.Infof("metric is off, clientId=%s", dcmp.client.GetClientID())
		dcmp.clientMeter = NewDefaultClientMeter(nil, false, nil, dcmp.client.GetClientID()) // DATA RACE
		return
	}

==================
==================

// github.com/apache/rocketmq-clients/golang@v0.0.0-20221125070211-5ba0ba13e773/producer_options.go:117
func (ps *producerSettings) IsValidateMessageType() bool {
[...]

PR #411, opened 2023-03-20
Fix: increase wait time in recovery tests to avoid flaking

### LINKED ISSUE #410: Flaky Test: TestRestoreDefaultClientSessionOneError
image

Anakin100100:
Currently the tests for recovering after TelemetryCommand transmission failure wait 3 seconds after the initial setup for the telemetry client goroutine to run. This is at the edge of expected wait time for the goroutine to go through all of the required code. Increasing the wait time for five seconds guarantees that by the time that we check the expected logs all of the testing scenario will have already completed. 

Fixes #410

### DIFF
--- a/.github/workflows/golang_build.yml
+++ b/.github/workflows/golang_build.yml
@@ -18,4 +18,7 @@ jobs:
           go-version: ${{ matrix.go }}
       - name: Build
         working-directory: ./golang
-        run: go build && go test ./...
+        run: go build
+      - name: Test
+        working-directory: ./golang
+        run: go test -v
--- a/.github/workflows/golang_build.yml
+++ b/.github/workflows/golang_build.yml
@@ -1,5 +1,5 @@
 name: Golang Build
-on: 
+on:
   workflow_call:
 jobs:
   build:
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-20.04, macos-11, windows-2022]
-        go: [1.17]
+        go: [1.19.4]
     steps:
       - name: Checkout
         uses: actions/checkout@v2
--- a/.github/workflows/golang_build.yml
+++ b/.github/workflows/golang_build.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-20.04, macos-11, windows-2022]
-        go: [1.19.4]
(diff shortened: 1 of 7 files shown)


guyinyou (approved):

Anakin100100:
It seems like this issue is not related to the wait time of the consumer goroutine. I'm going to try to reproduce in on Windows

aaron-ai:
@Anakin100100 Thank you for your contribution! I am not familiar with Golang, and I have a question: does Golang have a tool similar to Java's awaitility that can help us avoid using the sleep method and increase the speed of test execution?

Anakin100100:
@aaron-ai I'm not aware that such tool exists @guyinyou is something like that typical in Go? In this case skipping the sleep period is a bad idea because the function that we are testing is running possibly on another process as a goroutine so there is no guarantee of execution order. This would result in flaking. Running the tests in parallel is a viable idea. I will look into that.

Anakin100100:
I kinda suspect this could be a deadlock somewhere that is hit more often on windows maybe due to different goroutine implementation. I will try to use go 1.17 locally just to make sure.

guyinyou (approved):

Anakin100100:
I managed to reproduce the issue only once on a completely fresh installation of Go on my desktop. Subsequent builds and test runs run fine.

Anakin100100:
@aaron-ai could you give me access to run GitHub actions in my pull requests? Currently I can't see if the new changes fix the build issue in cicd. (They seem to work locally)

aaron-ai:
> @aaron-ai could you give me access to run GitHub actions in my pull requests? Currently I can't see if the new changes fix the build issue in cicd. (They seem to work locally)

I am trying to investigate the Github Action approval requirement of ASF project, it is a little bit different from other GitHub project.

Anakin100100:
@aaron-ai thanks for pointing out the version change. I think we can merge this now that the tests run fine. To sum up in this pr we:
 - corrected 2 flaky tests
 - fixed flaking on windows in cicd
 - corrected some cosmetic changes pointed out by staticcheck. 

I'm going to work on the data races in #294 next.

aaron-ai:
@Anakin100100  Could you comment on this issue https://github.com/apache/rocketmq-clients/issues/294? then I can assign it to you.

PR #461, opened 2023-04-04
chore(proto): introduce rocketmq-apis as submodule

### LINKED ISSUE #460: Introduce rocketmq-apis as submodule
Currently, we must manually import rocketmq protos every time there are changes. I propose adding rocketmq-apis repo as a submodule. So that we can receive updates more conveniently and less error-prone.

ShadowySpirits:
fix #460


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "rocketmq-apis"]
```
aaron-ai, line 1: Although we have placed IDL in [rocketmq-apis](https://github.com/apache/rocketmq-apis), I still don't think that rocketmq-apis is an appropriate name to describe IDL. "apis" seems more like a description for the interfaces of concrete code implementation. Therefore, it might be more fitting to place the submodule in a directory called "proto" or "IDL"?

aaron-ai:
Great job, it looks solid overall, and I left a comment.

.gitmodules
ShadowySpirits: Agree with your insight. I renamed the submodule to `protos`.

Issue #468, opened 2023-04-06
golang clinet does not implement ForwardMessageToDeadLetterQueue


Samuel-Solis-g:
Why doesn't golang implement ForwardMessageToDeadLetterQueue, is there any other way to implement early end retry on the client side and put the message into DLQ?

aaron-ai:
@Samuel-Solis-g Thank you for your interest. Please note that `ForwardMessageToDeadLetterQueue` is not a generic RPC request for consumers and is only relevant for PushConsumers.

Currently, there is no corresponding implementation of PushConsumer in the Go language, but some developers are working on it. It would be great if you could also be involved. see #463

PR #469, opened 2023-04-06
feat: golang client implementation ForwardMessageToDeadLetterQueue

tianping526:


tianping526:
fix: #468

golang/simple_consumer.go
```diff
@@ -183,6 +184,19 @@ func (sc *defaultSimpleConsumer) wrapAckMessageRequest(messageView *MessageView)
 	}
 }
 
+func (sc *defaultSimpleConsumer) wrapForwardMessageToDeadLetterQueueRequest(messageView *MessageView) *v2.ForwardMessageToDeadLetterQueueRequest {
```
aaron-ai, line 187: There should not be the `ForwardMessageToDeadLetterQueueRequest` in SimpleConsumer.

golang/simple_consumer.go
```diff
@@ -427,3 +441,27 @@ func (sc *defaultSimpleConsumer) Ack(ctx context.Context, messageView *MessageVi
 	sc.cli.doAfter(MessageHookPoints_ACK, messageCommons, duration, messageHookPointsStatus)
 	return nil
 }
+
+func (sc *defaultSimpleConsumer) ForwardMessageToDeadLetterQueue(ctx context.Context, messageView *MessageView) error {
```
aaron-ai, line 445: ditto

golang/simple_consumer.go
tianping526: Is there any good way to customize the retry policy? For example, if there are two messages m and n in topic A, I want to retry 3 times for m, but 100 times for n, and the retry interval can be decided by the client

PR #476, opened 2023-04-16
feat(rust): implement simple consumer

### LINKED ISSUE #475: Implement simple consumer for rust client
start rocketmq server and run the example as below:

```sh
cargo run --example simple_consumer
```

image

ShadowySpirits:
close #475 

Brief changelog:

1. implement heartbeat and telemetry rpc for both producer and simple consumer
2. implement receive and ack message rpc for simple consumer
3. add util module to abstract common functions
4. add simple consumer example and some tests

### DIFF
--- a/rust/README.md
+++ b/rust/README.md
@@ -14,7 +14,7 @@ Here are some preparations you may need to know (or refer to [here](https://rock
 
 1. rust and cargo
 2. protoc 3.15.0+
-3. Setup name server, broker, and [proxy](https://github.com/apache/rocketmq/tree/develop/proxy).
+3. setup name server, broker, and [proxy](https://github.com/apache/rocketmq/tree/develop/proxy).
 
 ### Run Example
 
--- a/rust/README.md
+++ b/rust/README.md
@@ -1,4 +1,4 @@
-# The Java Implementation of Apache RocketMQ Client
+# The Rust Implementation of Apache RocketMQ Client
 
 [RocketMQ Website](https://rocketmq.apache.org/)
 
--- a/rust/README.md
+++ b/rust/README.md
@@ -1,0 +1,29 @@
+# The Java Implementation of Apache RocketMQ Client
+
+[RocketMQ Website](https://rocketmq.apache.org/)
+
+## Overview
+
+Here is the rust implementation of the client for [Apache RocketMQ](https://rocketmq.apache.org/). Different from the [remoting-based client](https://github.com/apache/rocketmq/tree/develop/client), the current implementation is based on separating architecture for computing and storage, which is the more recommended way to access the RocketMQ service.
+
+Here are some preparations you may need to know (or refer to [here](https://rocketmq.apache.org/docs/quickStart/02quickstart)).
+
+## Getting Started
+
+### Requirements
+
+1. rust and cargo
+2. protoc 3.15.0+
+3. Setup name server, broker, and [proxy](https://github.com/apache/rocketmq/tree/develop/proxy).
+
+### Run Example
+
+Run the following command to start the example:
+
+```sh
+# send message via producer
+cargo run --example producer
+
+# consume message via simple consumer
+cargo run --example simple_consumer
+```
--- a/rust/src/conf.rs
+++ b/rust/src/conf.rs
@@ -14,31 +14,35 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+use crate::model::common::ClientType;
+use std::time::Duration;
+
 #[derive(Debug, Clone)]
 pub struct ClientOption {
-    name_space: String,
-    access_url: String,
-    enable_tls: bool,
+    pub(crate) client_type: ClientType,
+    pub(crate) group: String,
+    pub(crate) namespace: String,
+    pub(crate) access_url: String,
+    pub(crate) enable_tls: bool,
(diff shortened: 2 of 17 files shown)


lizhanhui:
If having protoc version >= 3.12, need to add option `--experimental_allow_proto3_optional` https://docs.rs/tonic-build/0.9.1/tonic_build/struct.Builder.html#method.protoc_arg

rust/src/conf.rs
```diff
@@ -62,31 +80,109 @@ pub enum LoggingFormat {
...
         &self.logging_format
     }
     pub fn set_logging_format(&mut self, logging_format: LoggingFormat) {
         self.logging_format = logging_format;
     }
 
+    pub fn consumer_group(&self) -> &str {
+        &self.consumer_group
+    }
+    pub fn set_consumer_group(&mut self, consumer_group: String) {
```
lizhanhui, line 182: Should accept AsRef<str>. That is, the signature should be 

```
pub fn set_consumer_group(&mut self, consumer_group: AsRef<str>)
```

This way, application developers can write 

```
let mut options = SimpleConsumerOption::default();
options.set_consumer_group("Group1");
```

lizhanhui: It's the same for other application developer-facing APIs.

ShadowySpirits:
> If having protoc version >= 3.12, need to add option `--experimental_allow_proto3_optional` https://docs.rs/tonic-build/0.9.1/tonic_build/struct.Builder.html#method.protoc_arg

We have no need to add this flag, protoc support optional since 3.15.

Reference: https://github.com/protocolbuffers/protobuf/blob/main/docs/field_presence.md#protoc-invocation

rust/src/conf.rs
ShadowySpirits: How about using `Into<String>` instead of `String`?

lizhanhui: Into<String> is also OK.

lizhanhui:
> 



> > If having protoc version >= 3.12, need to add option `--experimental_allow_proto3_optional` https://docs.rs/tonic-build/0.9.1/tonic_build/struct.Builder.html#method.protoc_arg
> 
> We have no need to add this flag, protoc support optional since 3.15.
> 
> Reference: https://github.com/protocolbuffers/protobuf/blob/main/docs/field_presence.md#protoc-invocation

If protoc prior to 3.15 should fail, we should check and make it fail explicitly.

ShadowySpirits:
> > 
> 
> > > If having protoc version >= 3.12, need to add option `--experimental_allow_proto3_optional` https://docs.rs/tonic-build/0.9.1/tonic_build/struct.Builder.html#method.protoc_arg
> > 
> > 
> > We have no need to add this flag, protoc support optional since 3.15.
> > Reference: https://github.com/protocolbuffers/protobuf/blob/main/docs/field_presence.md#protoc-invocation
> 
> If protoc prior to 3.15 should fail, we should check and make it fail explicitly.

I'll do this check, and add this requirement to the readme.

lizhanhui (approved):

rust/src/conf.rs
```diff
@@ -14,36 +14,40 @@
...
+    pub(crate) long_polling_timeout: Duration,
 }
 
 impl Default for ClientOption {
     fn default() -> Self {
         ClientOption {
-            name_space: "".to_string(),
+            client_type: ClientType::Producer,
+            group: "".to_string(),
+            namespace: "".to_string(),
```
aaron-ai, line 36: We currently have not exposed the concept of namespace to the client side, is it possible to remove the namespace here?

rust/src/conf.rs
```diff
@@ -14,36 +14,40 @@
...
             access_url: "localhost:8081".to_string(),
             enable_tls: false,
```
aaron-ai, line 38: Can we assume that SSL is enabled by default?

rust/src/conf.rs
```diff
@@ -62,31 +80,109 @@ pub enum LoggingFormat {
...
+
+impl Default for SimpleConsumerOption {
+    fn default() -> Self {
+        SimpleConsumerOption {
+            logging_format: LoggingFormat::Terminal,
+            consumer_group: "".to_string(),
+            prefetch_route: true,
+            topics: None,
+            namespace: "".to_string(),
+            await_duration: Duration::from_secs(1),
```
aaron-ai, line 166: Is it possible to extend the default time since the minimum time allowed by the server is 5s?

rust/src/conf.rs
aaron-ai: Producer group makes no sense in 5.x series?

rust/src/conf.rs
ShadowySpirits: Sure, but tls support is not implemented now. I will work on this later.

rust/src/conf.rs
ShadowySpirits: Which default value do you consider more suitable?

rust/README.md
```diff
@@ -0,0 +1,29 @@
+# The Java Implementation of Apache RocketMQ Client
```
aaron-ai, line 1: The Java Implementation => The Rust Implementation

aaron-ai (approved):

PR #485, opened 2023-04-19
doc(rust): add rust doc

### LINKED ISSUE #483: Improve test and doc for rust sdk
Task

- [ ] mock underlying component #482
- [ ] add test for util, conf, producer, and simple_consumer module
- [ ] add more doc

ShadowySpirits:
close #483 

Add rust doc for rust SDK: [Preview link](https://blog-git-previewrustdoc-sspirits.vercel.app/doc/rocketmq/index.html)

@lizhanhui @aaron-ai PTAL, you can login preview website via your Github account and leave comments.

image

After this pr, we can publish the rust SDK on crates.io.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -18,6 +18,17 @@
 name = "rocketmq"
 version = "0.1.0"
 edition = "2021"
+authors = [
+    "SSpirits <admin@lv5.moe>",
+    "Zhanhui Li <lizhanhui@gmail.com>",
+]
+
+license = "MIT/Apache-2.0"
+readme = "./README.md"
+repository = "https://github.com/apache/rocketmq-clients"
+documentation = "https://docs.rs/rocketmq"
+description = "Rust client for Apache RocketMQ"
+keywords = ["rocketmq", "api", "client", "sdk", "grpc"]
 
 [dependencies]
 tokio = { version = "1", features = ["full"] }
--- a/rust/src/conf.rs
+++ b/rust/src/conf.rs
@@ -14,9 +14,13 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+//! Configuration of RocketMQ rust client.
+
 use crate::model::common::ClientType;
 use std::time::Duration;
 
+/// `ClientOption` is the configuration of internal `Client`, which manages the connection and request with RocketMQ proxy.
 #[derive(Debug, Clone)]
 pub struct ClientOption {
     pub(crate) client_type: ClientType,
@@ -46,6 +50,7 @@ impl ClientOption {
     pub fn access_url(&self) -> &str {
         &self.access_url
     }
+    /// set the access url of RocketMQ proxy
     pub fn set_access_url(&mut self, access_url: impl Into<String>) {
         self.access_url = access_url.into();
     }
@@ -53,6 +58,7 @@ impl ClientOption {
     pub fn enable_tls(&self) -> bool {
         self.enable_tls
     }
+    /// set whether to enable tls, default is true
     pub fn set_enable_tls(&mut self, enable_tls: bool) {
         self.enable_tls = enable_tls;
     }
@@ -60,6 +66,7 @@ impl ClientOption {
     pub fn timeout(&self) -> &Duration {
         &self.timeout
     }
+    /// set the timeout of connection and generic request, default is 3 seconds
     pub fn set_timeout(&mut self, timeout: Duration) {
         self.timeout = timeout;
     }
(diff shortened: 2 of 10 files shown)


rust/Cargo.toml
```diff
@@ -18,6 +18,17 @@
 name = "rocketmq"
```
aaron-ai, line 18: We currently do not specify the version of the Rust toolchain in the workflow of the Github Action. In fact, the Rust version used in the default runner environment is relatively high(see https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#rust-tools), so compilation is always no problem. However, I did encounter compatibility issues when compiling under the 1.63 version of the Rust toolchain, and the problem was not solved until I switched to 1.68.2.

I think it would be best to specify the [rust-version](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field) in `Cargo.toml` to determine the minimum version of the Rust toolchain required(e.g. https://github.com/assert-rs/predicates-rs/blob/660f658285702e0219d9d76e65881ea5b8536a2c/Cargo.toml#L8).

In addition, could you help to make it clear in the documentation and GitHub Action workflow what our version requirements are?

ShadowySpirits: I have set `edition = "2021"` in `Cargo.toml`. It limits the minimum version of rust to 1.56.0.

Reference: https://doc.rust-lang.org/edition-guide/rust-2021/index.html

aaron-ai: Nice! I think it is a great choice for a Rust lib to have a low requirement on the version of Rust toolchain.

Meanwhile, `edition` and `rust-version` seem to have different semantics. The former focuses on the major features of Rust, while the latter is directly facing the compiler. So perhaps they cannot be replaced with each other? We also need to make it clear to our developers what is the minimum version required for current client build. From my current attempts, version 1.56 does not guarantee the success of project build.

see: https://github.com/aaron-ai/rocketmq-clients/actions/runs/4740564022/jobs/8416534825

ShadowySpirits: I have set `rust-version` to `1.61` which is this crate's msrv and tested it on my Mac.

image

aaron-ai: Fantastic work, our client now has an MSRV! Looking forward to you configuring the MSRV in our Github Action workflow in the next pull request to avoid potential compatibility issues that may be inadvertently introduced during the community development process.

aaron-ai (approved):

rust/src/conf.rs
```diff
@@ -46,27 +50,33 @@ impl ClientOption {
     pub fn access_url(&self) -> &str {
         &self.access_url
     }
+    /// set the access url of RocketMQ proxy
```
lizhanhui, line 53: Please follow rust documentation examples to include arguments
https://doc.rust-lang.org/rust-by-example/meta/doc.html

rust/src/model/message.rs
```diff
@@ -267,46 +292,57 @@ impl MessageView {
...
         &self.properties
     }
 
+    /// get message group of fifo message
     pub fn message_group(&self) -> Option<&str> {
         self.message_group.as_deref()
     }
 
+    /// get delivery timestamp of delay message
     pub fn delivery_timestamp(&self) -> Option<i64> {
```
lizhanhui, line 331: Return Option<Instant>?

rust/src/model/message.rs
```diff
@@ -267,46 +292,57 @@ impl MessageView {
...
         self.delivery_timestamp
     }
 
+    /// get born host of message
     pub fn born_host(&self) -> &str {
         &self.born_host
     }
 
+    /// get born timestamp of message
     pub fn born_timestamp(&self) -> i64 {
```
lizhanhui, line 341: Same here

rust/src/model/message.rs
```diff
@@ -267,46 +292,57 @@ impl MessageView {
...
         self.born_timestamp
     }
 
+    /// get delivery attempt of message
     pub fn delivery_attempt(&self) -> i32 {
```
lizhanhui, line 346: i32 or usize?

rust/src/model/message.rs
ShadowySpirits: Instant seems could not be created by ourselves.

> A measurement of a monotonically nondecreasing clock. Opaque and useful only with [Duration](https://doc.rust-lang.org/std/time/struct.Duration.html).

lizhanhui: OK

Issue #459, opened 2023-04-04
Enhance Rust SDK CI Pipeline


lizhanhui:
Jon is a very professional and popular Rust developer... We should reuse pipelines shared: https://github.com/jonhoo/rust-ci-conf

lizhanhui:
@ShadowySpirits Would you like to take a look?

ShadowySpirits:
> @ShadowySpirits Would you like to take a look?

Sure, please assign it to me.

ShadowySpirits:
I have added test coverage ci for rust SDK #489

PR #490, opened 2023-04-21
ci(rust): add check ci

ShadowySpirits:
link to #459


aaron-ai (commented):
Can you collect the Rust related checks into the job of rust-build, since build-result collects the results of all language builds and is required? If the Rust checks can be classified into rust-build, then as long as build-result is successful, it indicates that all Rust checks have been successful.

aaron-ai:
> Can you collect the Rust related checks into the job of rust-build, since build-result collects the results of all language builds and is required? If the Rust checks can be classified into rust-build, then as long as build-result is successful, it indicates that all Rust checks have been successful.

build-result collects all results: https://github.com/apache/rocketmq-clients/actions/runs/4763211461?pr=490

image

https://github.com/apache/rocketmq-clients/actions/runs/4763211432?pr=490

image

ShadowySpirits:
> https://github.com/apache/rocketmq-clients/actions/runs/4763211461?pr=490

IMO, code style check is different from rust-build. Additionally, rust-build will be processed in the different platforms which are meaningless for the code check.

aaron-ai:
You are right, general checks do not usually require cross-platform operation systems.

Meanwhile, we can actually specify these checks to a single or other platforms, referring to [opentelemetry-instrumentation-compatibility](https://github.com/apache/rocketmq-clients/blob/585f93e740098f23a5873499438e1876fe7cb89e/.github/workflows/java_build.yml#L25). see: https://github.com/apache/rocketmq-clients/actions/runs/4668191248?pr=472

ShadowySpirits:
> > > https://github.com/apache/rocketmq-clients/actions/runs/4763211461?pr=490
> > 
> > 
> > IMO, code style check is different from rust-build. Additionally, rust-build will be processed in the different platforms which are meaningless for the code check.
> 
> Sorry, I think my previous expression was not accurate, confusing the job and workflow files. You can refer to the latest comments.

I have merged rust-check and rust-build. The build job depends on fmt check and msrv check for now.

.github/workflows/rust_build.yml
```diff
@@ -2,27 +2,114 @@ name: Rust Build
...
+        msrv: [1.61]
+    name: msrv / ${{ matrix.msrv }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: true
+      - name: Install ${{ matrix.msrv }}
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ matrix.msrv }}
```
aaron-ai, line 79: Do we need to set [override](https://github.com/actions-rs/toolchain#example-workflow:~:text=false-,override,-Set%20installed%20toolchain) to true? I am not sure if the additional version we installed will be enabled if we do not set it to true.

ShadowySpirits: I manually specified the version used by `cargo check`

image

.github/workflows/rust_build.yml
```diff
@@ -2,27 +2,114 @@ name: Rust Build
...
+      - name: Install protoc
+        uses: arduino/setup-protoc@v1
+        with:
+          version: '3.x'
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Check MSRV ${{ matrix.msrv }}
+        working-directory: ./rust
+        run: cp Cargo.minimal.lock Cargo.lock && cargo +${{ matrix.msrv }} check --locked
```
aaron-ai, line 87: I'm not quite sure why we need to introduce Cargo.lock, and I currently find that one of the major reasons why it can't be compiled on 1.61 is due to the [predicates-rs](https://github.com/assert-rs/predicates-rs) package. I wonder if we can ensure compatibility of compilation by restricting the version of this package.

Below is a configuration that I have verified, for your reference:

```toml
predicates = "< 3.0.0"
predicates-tree = "< 1.0.8"
predicates-core = "< 1.0.6"
```

ShadowySpirits: Our sdk lib is integrated by other application, so that pin the version of dependencies is not a good choice.

aaron-ai (approved):

PR #541, opened 2023-06-12
Fix cpp SimpleConsumer

yulangz:
Fix the bug that SimpleConsumer in the cpp version cannot obtain the broker Routing table.

### DIFF
--- a/cpp/source/client/ClientManagerImpl.cpp
+++ b/cpp/source/client/ClientManagerImpl.cpp
@@ -24,12 +24,9 @@
 #include <utility>
 #include <vector>
 
-#include "apache/rocketmq/v2/definition.pb.h"
 #include "InvocationContext.h"
 #include "LogInterceptor.h"
 #include "LogInterceptorFactory.h"
-#include "rocketmq/Logger.h"
-#include "spdlog/spdlog.h"
 #include "MessageExt.h"
 #include "MetadataConstants.h"
 #include "MixAll.h"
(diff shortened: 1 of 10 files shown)


lizhanhui:
Please format your pull-request changes using clang-format

cpp/source/client/ClientManagerImpl.cpp
```diff
@@ -764,7 +764,7 @@ State ClientManagerImpl::state() const {
 }
 
 MessageConstSharedPtr ClientManagerImpl::wrapMessage(const rmq::Message& item) {
-  assert(item.topic().resource_namespace() == resource_namespace_);
+  // assert(item.topic().resource_namespace() == resource_namespace_);
```
lizhanhui, line 767: Why comment out this assertion?

cpp/source/rocketmq/Producer.cpp
```diff
@@ -87,7 +87,8 @@ ProducerBuilder& ProducerBuilder::withConfiguration(Configuration configuration)
 }
 
 ProducerBuilder& ProducerBuilder::withTopics(const std::vector<std::string>& topics) {
-  impl_->topicsOfInterest(topics);
+  auto topics_copy = topics;
```
lizhanhui, line 90: I do not see how the proposed change is superior to the original. Any reason for this change?

yulangz: 1. The input parameter of `ProducerBuilder::withTopics` is a const reference, while the input parameter of `topicsOfInterest` is not a constant reference. The original approach cannot be compiled.

2. The internal implementation of `topicsOfInterest` will use `std:: vector<>. swap()`, which will cause the internal memory space held by the incoming `topics` to be swapped out, which does not comply with the semantics of read-only input parameters for `withTopics`.

cpp/source/client/ClientManagerImpl.cpp
yulangz: The resource_namespace_ is always "", this causes this assert to fail.

cpp/source/rocketmq/Producer.cpp
lizhanhui: > The input parameter of ProducerBuilder::withTopics is a const reference, while the input parameter of topicsOfInterest is not a constant reference. The original approach cannot be compiled.
No, this is not true.  `ProducerImpl::topicsOfInterest(std::vector<std::string> topics)` copy happens implicitly.

> 2. The internal implementation of topicsOfInterest will use std:: vector<>. swap(), which will cause the internal memory space held by the incoming topics to be swapped out, which does not comply with the semantics of read-only input parameters for withTopics.
No, it is using the implicit copy of the topics.

cpp/source/client/ClientManagerImpl.cpp
lizhanhui: @yulangz It's true `resource_namespace_` is empty for the moment. But the assertion should pass

cpp/source/rocketmq/Producer.cpp
lizhanhui: @yulangz

lizhanhui (changes requested):
The pull request understands C++ inaccurate. See this example: https://onecompiler.com/cpp/3zpv8g247

PR #544, opened 2023-06-12
Polish the cpp examples

yulangz:
Add `withTopics` in cpp examples.
If this is not enabled, it will cause the connection between the client and the server to fail.


lizhanhui:
> Add `withTopics` in cpp examples. If this is not enabled, it will cause the connection between the client and the server to fail.

`withTopics` should be optional by design and there should be no error in case of omission.

lizhanhui (approved):

PR #580, opened 2023-08-03
golang: optimize the logic of obtaining routing tasks

guyinyou:
optimize the logic of obtaining routing tasks


xdkxlk (approved):

leizhiyuan:
After running for a period of time, the consumer may encounter a timeout in the request routing table due to network fluctuations or other abnormal reasons. In previous versions, an error message "messageQueues is empty" would occur and it seems unable to recover. If you encounter this issue, please upgrade to this version.

PR #584, opened 2023-08-07
fix(doc): correct incorrect statements in Design and Workflow sections

Gezi-lzq:
### Brief Description

1. The client identifier "macbook-pro@90009@0@2dyeb8lep" can be divided into four parts using the separator "@".Among these parts, the client index within the current process should be 0.

2. The sentence "Check the cache is full, consumer will try to receive message immediately, otherwise retry after 1 second" is expressed incorrectly. It should be stated that when the cache is full, the consumer should retry after 1 second.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -55,7 +55,7 @@ Client identifier provides identity information for each client event within the
 
 * `macbook-pro`: hostname of device;
 * `90009`: process identifier;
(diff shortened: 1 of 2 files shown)


aaron-ai (approved):
Nice catch!

aaron-ai (approved):
Nice catch!

Issue #594, opened 2023-09-02
[Feature] csharp: replace nlog with Microsoft.Extensions.Logging


eventhorizon-cli:
### Programming Language of the Client

C#

### Is Your Feature Request Related to a Problem?

The logs of csharp client are written to files by default, and cannot be written only to the console, this is not very friendly to the .net developers.

1. This kind of logging is not in line with the normal practice of .NET developers.
2. If it is a product environment, long-term log file squeezes can cause certain issues.
3. The control of log output is not flexible enough.

### Describe the Solution You'd Like

The more commonly used logging system in .net sdk is  Microsoft.Extensions.Logging.
https://learn.microsoft.com/en-us/dotnet/core/extensions/logging?tabs=command-line

The sdk can expose an API for configuring LoggerFactory, allowing users to configure it freely.

The design of EF Core can be referenced.
https://learn.microsoft.com/en-us/ef/core/logging-events-diagnostics/extensions-logging?tabs=v3#other-application-types

I tried to modify the logger in rocketmq client. Here is what it looks like after modification.

```C#
public static class MqLogManager
{
    private static ILoggerFactory _loggerFactory;

    public static ILogger<T> CreateLogger<T>()
    {
        return _loggerFactory?.CreateLogger<T>() ?? NullLogger<T>.Instance;
    }
    
    public static ILogger CreateLogger(string categoryName)
    {
        return _loggerFactory?.CreateLogger(categoryName) ?? NullLogger.Instance;
    }

    public static void UseLoggerFactory(ILoggerFactory loggerFactory)
    {
        _loggerFactory = loggerFactory ?? throw new ArgumentNullException(nameof(loggerFactory));
    }
}
```

usages:

```C#
MqLogManager.UseLoggerFactory(LoggerFactory.Create(builder =>
{
    builder.AddConsole();
}));

public abstract class Client
{
    private static readonly ILogger Logger = MqLogManager.CreateLogger<Client>();

    protected virtual async Task Start()
    {
        Logger.LogDebug($"Begin to start the rocketmq client, clientId={ClientId}");
    }
}
```

If possible, I would like to submit a PR, thanks.

### Describe Alternatives You've Considered

Modify the current nlog code to provide the option not to output to a file

PR #596, opened 2023-09-05
replace nlog with Microsoft.Extensions.Logging

eventhorizon-cli:
replace nlog with Microsoft.Extensions.Logging https://github.com/apache/rocketmq-clients/issues/594

With the `Microsoft.Extensions.Logging`, the sdk can expose an API for configuring LoggerFactory, allowing users to configure logging system freely.


lizhanhui:
@eventhorizon-cli CI found a few format issues

eventhorizon-cli:
md format issues have been fixed, thanks @lizhanhui

aaron-ai:
Thank you for your suggestion. I think replacing the logging framework is quite reasonable, and more discussion is needed regarding changing the default behavior of log output.

lizhanhui:
> If necessary, an nlog LoggerFactory can configured as the default LoggerFactory to output logs in the current format. But users are allowed to configure their LoggerFactory. @aaron-ai

This sounds reasonable. It would be nice not to change the default behavior in general.

lizhanhui (approved):

PR #623, opened 2023-10-15
feat(rust): support changing invisible duration

ShadowySpirits:
Brief changelog for rust client:

1. Support changing invisible duration RPC
2. Suppress the warning about no settings telemetry command handler


rust/examples/simple_consumer.rs
```diff
@@ -72,6 +74,19 @@ async fn main() {
                 ack_result.unwrap_err()
             );
         }
+
```
lizhanhui, line 77: How about controlling this through a feature, instead of commenting

lizhanhui:
And there are a few clippy warnings to fix

PR #636, opened 2023-11-09
Replace lazy_static with once_cell

ixmoyren:
lazy_static is almost inactive, most of the projects in the rust community use once_cell instead of lazy_static, while once_cell's performance is  [a little bit more](https://github.com/matklad/once_cell/tree/master/examples) and the development is active.

1. remove lazy_static dependency
2. upgrade once_cell


lizhanhui (dismissed):
Looks good

lizhanhui (approved):

PR #665, opened 2024-01-04
[JAVA]only readable message queues are allowed to be assigned to push consumer

glcrazier:

When PushConsumer gets new assignments from the server, it does not check permission of message queues.

### DIFF
--- a/java/client/src/main/java/org/apache/rocketmq/client/java/impl/consumer/PushConsumerImpl.java
+++ b/java/client/src/main/java/org/apache/rocketmq/client/java/impl/consumer/PushConsumerImpl.java
@@ -34,6 +34,7 @@ import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.MoreExecutors;
 import java.time.Duration;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


java/client/src/main/java/org/apache/rocketmq/client/java/impl/consumer/PushConsumerImpl.java
```diff
@@ -389,6 +390,14 @@ public void onSuccess(Assignments latest) {
...
                             log.info("Attention!!! acquired empty assignments from remote, but existed assignments"
                                 + " is not empty, topic={}, clientId={}", topic, clientId);
+                        } else {
+                            List<Assignment> newAssignmentList = new ArrayList<>(latest.getAssignmentList().size());
+                            for (Assignment assignment : latest.getAssignmentList()) {
+                                if (assignment.getMessageQueue().getPermission().isReadable()) {
+                                    newAssignmentList.add(assignment);
+                                }
+                            }
+                            latest = new Assignments(newAssignmentList);
```
aaron-ai, line 400: Add more tests here?

glcrazier: sure, will do.

aaron-ai:
The queue returned by the server must always be readable, which is actually an agreement established with the server side. 

@drpmma please review whether this modification is necessary, thanks.

PR #680, opened 2024-02-02
[ISSUE #679] Fix judgment of topic route equality and optimize loadBalancer

### LINKED ISSUE #679: [Bug] Go client can't receive certain broker's messages

- [X] I have searched the [GitHub Issues](https://github.com/apache/rocketmq-clients/issues) and [GitHub Discussions](https://github.com/apache/rocketmq-clients/discussions)  of this repository and believe that this is not a duplicate.

Go

centos

master

### Run or Compiler Version

_No response_

### Describe the Bug

We have a broker cluster with 4 master nodes, and use go client simple consumer to consume. But we found that messages of broker-0 can't be received.

### Steps to Reproduce

1. Start up a multi-master cluster
2. Start up go client simple consumer, setting awaitDuration=20s
3. Produces certain messages.

Receive all brokers' messages.

### What Did You See Instead?

Messages of certain broker can't be received.

### Additional Context

image
image

原因应该是在定时更新 topic 路由这里，判断路由是否变化时使用了 reflect.DeepEqual。因为 pb message 会包含一些额外的内部字段（比如 sizeCache），所以会误判成路由发生变化，就会导致 subscriptionLoadBalancer 的 index 每 30s 就会被重置，如果 broker 多节点且 pollingTime 比较长的话，就会出现某些 broker 一直不能被选中的情况。

redlsz:

Fixes #679 

### Brief Description

1. correct judgment of topic route equality
2. do not reset index when update loadBalancer

### DIFF
--- a/golang/client.go
+++ b/golang/client.go
@@ -504,16 +504,19 @@ func (cli *defaultClient) startUp() error {
 	f := func() {
 		cli.router.Range(func(k, v interface{}) bool {
 			topic := k.(string)
-			oldRoute := v
 			newRoute, err := cli.queryRoute(context.TODO(), topic, cli.opts.timeout)
 			if err != nil {
 				cli.log.Errorf("scheduled queryRoute err=%v", err)
 			}
-			if newRoute == nil && oldRoute != nil {
+			if newRoute == nil && v != nil {
 				cli.log.Info("newRoute is nil, but oldRoute is not. do not update")
 				return true
 			}
-			if oldRoute == nil && len(newRoute) > 0 || oldRoute != nil && !routeEqual(oldRoute.([]*v2.MessageQueue), newRoute) {
+			var oldRoute []*v2.MessageQueue
+			if v != nil {
+				oldRoute = v.([]*v2.MessageQueue)
+			}
+			if !routeEqual(oldRoute, newRoute) {
 				cli.router.Store(k, newRoute)
 				switch impl := cli.clientImpl.(type) {
 				case *defaultProducer:
--- a/golang/client.go
+++ b/golang/client.go
@@ -513,17 +513,6 @@ func (cli *defaultClient) startUp() error {
 				cli.log.Info("newRoute is nil, but oldRoute is not. do not update")
 				return true
 			}
-			routeEqual := func(old, new []*v2.MessageQueue) bool {
-				if len(old) != len(new) {
-					return false
-				}
-				for i := 0; i < len(old); i++ {
-					if !proto.Equal(old[i], new[i]) {
-						return false
-					}
-				}
-				return true
-			}
 			if oldRoute == nil && len(newRoute) > 0 || oldRoute != nil && !routeEqual(oldRoute.([]*v2.MessageQueue), newRoute) {
 				cli.router.Store(k, newRoute)
 				switch impl := cli.clientImpl.(type) {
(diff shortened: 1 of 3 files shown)


golang/client.go
```diff
@@ -513,18 +513,39 @@ func (cli *defaultClient) startUp() error {
 				cli.log.Info("newRoute is nil, but oldRoute is not. do not update")
 				return true
 			}
-			if !reflect.DeepEqual(newRoute, oldRoute) {
+			routeEqual := func(old, new []*v2.MessageQueue) bool {
```
lizhanhui, line 516: In C++ implementation, I used [MessageDifferencer](https://protobuf.dev/reference/cpp/api-docs/google.protobuf.util.message_differencer/) to achieve this; Golang has [protocmp](https://pkg.go.dev/google.golang.org/protobuf/testing/protocmp). Can we use protocmp to compare generically instead of manual work, which is error-prone over time?

redlsz: > In C++ implementation, I used [MessageDifferencer](https://protobuf.dev/reference/cpp/api-docs/google.protobuf.util.message_differencer/) to achieve this; Golang has [protocmp](https://pkg.go.dev/google.golang.org/protobuf/testing/protocmp). Can we use protocmp to compare generically instead of manual work, which is error-prone over time?

The [go-cmp](https://pkg.go.dev/github.com/google/go-cmp/cmp) and [protocmp](https://pkg.go.dev/google.golang.org/protobuf/testing/protocmp) seem being for testing. And manual comparison here is actually based on [proto.Equal](https://github.com/protocolbuffers/protobuf-go/blob/v1.32.0/proto/equal.go#L41)

leizhiyuan:
add some ref 

[protocmp](https://protobuf.dev/reference/go/faq/#deepequal)

it seems only [proto.Equal](https://pkg.go.dev/google.golang.org/protobuf/proto#Equal) can be used in production code 

where protocmp can be used in test (for best practice)

> Why does reflect.DeepEqual behave unexpectedly with protobuf messages?
> Generated protocol buffer message types include internal state which can vary even between equivalent messages.
> 
> In addition, the reflect.DeepEqual function is not aware of the semantics of protocol buffer messages, and can report differences where none exist. For example, a map field containing a nil map and one containing a zero-length, non-nil map are semantically equivalent, but will be reported as unequal by reflect.DeepEqual.
> 
> Use the [proto.Equal](https://pkg.go.dev/google.golang.org/protobuf/proto#Equal) function to compare message values.
> 
> In tests, you can also use the ["github.com/google/go-cmp/cmp"](https://pkg.go.dev/github.com/google/go-cmp/cmp?tab=doc) package with the [protocmp.Transform()](https://pkg.go.dev/google.golang.org/protobuf/testing/protocmp#Transform) option. The cmp package can compare arbitrary data structures, and [cmp.Diff](https://pkg.go.dev/github.com/google/go-cmp/cmp#Diff) produces human-readable reports of the differences between values.
> 
> if diff := cmp.Diff(a, b, protocmp.Transform()); diff != "" {
>   t.Errorf("unexpected difference:\n%v", diff)
> }

golang/client.go
lizhanhui: All right.

lizhanhui (changes requested):
A few follow-up issues to resolve:
1, Test cases are required to verify this change, which also ensures future changes will not break;
2, CI is broken

redlsz:
> A few follow-up issues to resolve: 1, Test cases are required to verify this change, which also ensures future changes will not break; 2, CI is broken

1. done
2. It seems to be not related to this issue.

lizhanhui (approved):
Nice work

PR #710, opened 2024-03-17
[rust] check route equality and keep existing route in case of failure

glcrazier:
…o query route from remote

Fixes #issue_id

### Brief Description

In current implementation, the route is updated each time after querying remote queues. This will cause the index to be reset to zero that some queues won't be selected for polling or sending messages if the route is updated frequently. And in case of failure route requests, the route would be removed instead of being kept for availability.

The commit fixed these two issues above.

add 2 test cases.


lizhanhui (approved):
Looks good

PR #735, opened 2024-04-26
修复当使用json文件日志的时候，panic的问题

fuyoo:
1.可以自定义配置文件路径
2.当配置文件创建失败时，退回到输出日志到stdout


lizhanhui:
We are building a library and the correct way to handling logging should be "Libraries should link only to the log crate, and use the provided macros to log whatever information will be useful to downstream consumers", cited from https://crates.io/crates/log

We should get rid of slog and use log crate only.

PR #767, opened 2024-06-09
[rust] implement pushconsumer

### LINKED ISSUE #683: [Feature] Implement PushConsumer for Rust SDK

Rust

We need to align features with Java/C++/...

N/A

N/A

### Additional Context

N/A

glcrazier:

Fixes #683 

### Brief Description

This is an implementation of standard and fifo PushConsumer which can consumes messages with listener provided. 

Sufficient unit test cases are provided and integration tests are executed.


lizhanhui:
Thanks @glcrazier for your nice work and I would join you and review this promising feature ASAP.

rust/src/push_consumer.rs
```diff
@@ -0,0 +1,1303 @@
...
+use crate::util::{build_endpoints_by_message_queue, build_push_consumer_settings};
+use crate::{log, pb};
+
+const OPERATION_NEW_PUSH_CONSUMER: &str = "push_consumer.new";
+const OPERATION_RECEIVE_MESSAGE: &str = "push_consumer.receive_message";
+const OPERATION_ACK_MESSAGE: &str = "push_consumer.ack_message";
+const OPERATION_START_PUSH_CONSUMER: &str = "push_consumer.start";
+const OPERATION_CHANGE_INVISIBLE_DURATION: &str = "push_consumer.change_invisible_duration";
+
+pub type MessageListener = Box<dyn Fn(&MessageView) -> ConsumeResult + Send + Sync>;
```
lizhanhui, line 50: Fn requires immutability, kind of excessively strict. Use FnMut?
Here is a good reference to read:
https://stackoverflow.com/questions/41081240/idiomatic-callbacks-in-rust

rust/src/push_consumer.rs
```diff
@@ -0,0 +1,1303 @@
...
+    async fn receive_messages<T: RPCClient + 'static>(
+        &mut self,
+        rpc_client: &mut T,
+        message_handler_queue: &mut VecDeque<MessageHandler>,
+        poll_tx: mpsc::Sender<()>,
+    ) {
+        let poll_result = self.poll_messages(rpc_client).await;
+        if let Ok(messages) = poll_result {
+            for message in messages {
+                let consume_result = (self.message_listener)(&message);
```
lizhanhui, line 418: On message arrival, we will consume the message in the current task? This would limit the parallelism of message processing to # of actors.

For concurrent model, in which out-of-order execution is allowed, we shall spawn a task for each mini-batch of messages, size of the mini-batch is up to configuration.

lizhanhui: For FIFO model, this is exactly what we need and we shall keep it as it is.

rust/src/push_consumer.rs
```diff
@@ -0,0 +1,1303 @@
...
+                match consume_result {
+                    ConsumeResult::SUCCESS => {
+                        let result = self.ack_message(rpc_client, &message).await;
+                        if result.is_err() {
```
lizhanhui, line 422: Error handling is not sufficient here.
For FIFO messages, we have to repeatedly retry ack the message before getting opportunity  to consume the next one.

For standard message model, we also need to retry because consume a message might be expensive in terms of resource consumption.

lizhanhui: Ah, you have implemented the retry mechanism.

rust/src/push_consumer.rs
```diff
@@ -0,0 +1,1303 @@
...
+                            info!(
```
lizhanhui, line 423: It would be better to define an abstract model here, some kind of ProcessTask and encapsulate process-ack-retry and order-keeping.
That is, an message queue actor has one(FIFO) or multiple(standard model) ProcessTask, one message arrival, put messages to existing ProcessTask(for FIFO) or spawn new ProcessTasks.

In the process task, we have the consume semantics implemented.

rust/src/push_consumer.rs
glcrazier: The FnMut requires exclusive access in threads which means incapable of concurrent consumption. Do you have any idea?

rust/src/push_consumer.rs
glcrazier: You idea is great, and I refactored that the consumption of standard and fifo is capsulated in separate implementation. And I also extract the acknowledgement work into a single implementation.

rust/src/push_consumer.rs
glcrazier: Now it's done

rust/src/push_consumer.rs
lizhanhui: Let's merge it ASAP and file a new pull request to improve.

lizhanhui (dismissed):

lizhanhui:
MSRV set to 1.61 does not make sense 
run cargo msrv, we are having 
image

.github/workflows/rust_build.yml
```diff
@@ -45,27 +52,26 @@ jobs:
...
         env:
           RUSTDOCFLAGS: --cfg docsrs
   msrv:
     runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./rust
     strategy:
       matrix:
-        msrv: [1.61]
```
ShadowySpirits, line 56: Rust version defined in Cargo.toml should be also bumped up to 1.70.0

lizhanhui (approved):
OK

Issue #667, opened 2024-01-15
[Bug] rocketmq-clients/golang与rocketmq-clients/java对数据解压的实现方式不一致导致gzip解压数据出错


Justin1210:
### Before Creating the Bug Report

- [X] I found a bug, not just asking a question, which should be created in [GitHub Discussions](https://github.com/apache/rocketmq-clients/discussions).

- [X] I have searched the [GitHub Issues](https://github.com/apache/rocketmq-clients/issues) and [GitHub Discussions](https://github.com/apache/rocketmq-clients/discussions)  of this repository and believe that this is not a duplicate.

- [X] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

### Programming Language of the Client

Go

### Runtime Platform Environment

linux

### RocketMQ Version of the Client/Server

消息队列 RocketMQ 版、RocketMQ5.1.4

### Run or Compiler Version

openjdk version "1.8.0_392"

### Describe the Bug

当消息生产者生产的消息时经过gzip压缩处理后，使用rocketmq-clients/golang 下的SimpleConsumer进行消费时无法正常解压出消息的body， 调试发现调用解压数据的操作utils.GZIPDecode会有错误消息：“gzip:invalid header”输出。我尝试使用rocketmq-clients/java下的SimpleConsumer例子测试，发现可以正常消费消息解压出数据，也就是说java版实现压缩和解压的方式与go的并不相同这个是为什么？java压缩使用的是zlib，在go中使用compress/gzip按标准的gzip解压，为什么不让它一致呢？
![WechatIMG17](https://github.com/apache/rocketmq-clients/assets/10846649/16a17580-1c93-4e1c-8f47-d4ad8ca6078f)

![WX20240115-180435@2x](https://github.com/apache/rocketmq-clients/assets/10846649/fbfab07b-eaba-4b9c-89ac-78f608edf55e)

### Steps to Reproduce

消息生产者使用rocketmq-client 4.8.0 去生产一条比较大的消息，消息的body大于compressMsgBodyOverHowmuch设定默认的值4K，然后久会触发压缩消息操作UtilAll.compress(body, this.zipCompressLevel)，然后使用rocketmq-clients/golang 下的SimpleConsumer去消费这条消息，就出现错误了

### What Did You Expect to See?

数据能正常解析出来

### What Did You See Instead?

[...]

PR #855, opened 2024-10-30
Resolve the issue of garbled characters in the Golang language message body GzipDecoder

dongpoge:
When the amount of data in the message body is relatively large, decompression may result in errors;
当客户发发送一个数据量大的消息体时，由于进行了gzip 压缩，golang 版本的代码，消费数据时，原先GZIPDecode 方法有问题，解压出来是乱码，优化后，可以正常消费和解压缩，测试没有问题

Fixes #issue_id

#667

### DIFF
--- a/golang/pkg/utils/utils.go
+++ b/golang/pkg/utils/utils.go
@@ -19,11 +19,11 @@ package utils
 
 import (
 	"bytes"
-	"compress/gzip"
+	"compress/zlib"
 	"context"
 	"encoding/hex"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"net"
 	"net/url"
 	"os"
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (commented):
### Summary

This PR changes `GZIPDecode` from `compress/gzip` to `compress/zlib` to fix garbled output for large message bodies. While the fix may be correct, there's a compatibility concern that needs clarification.

### Concerns

- **[Critical] Compression format mismatch** — The function is named `GZIPDecode` and the RocketMQ server/proxy sends gzip-compressed message bodies (per the protocol spec). Switching to `compress/zlib` would break decompression if the data is actually gzip-formatted. Could you confirm:
  1. What compression format does the server actually use for large message bodies?
  2. Was the original issue caused by the server sending zlib instead of gzip, or was there a different bug in the gzip reader usage?

- **[Warning] Function naming** — If the data is indeed zlib-compressed, the function should be renamed to `ZlibDecode` to avoid confusion. If it's gzip, the fix should stay with `compress/gzip`.

- **[Info] Error handling improvement** — The new error handling (`return nil, err` instead of `var out []byte; return out, err`) is cleaner. 👍

Could you share a test case or the actual compressed bytes that were failing? This would help verify the root cause.

---

PR #856, opened 2024-10-31
Fix Utilities.GetMacAddress may return wrong result

Executor-Cheng:
`Utilities.GetMacAddress` may return empty `byte[]` that crashes `MessageIdGenerator.MessageIdGenerator()` and `MessageIdGenerator`'s type initializer.

https://github.com/apache/rocketmq-clients/blob/33caa017045d6dc0339fc32d61d8f8c7169068af/csharp/rocketmq-client-csharp/MessageIdGenerator.cs#L46

### DIFF
--- a/csharp/rocketmq-client-csharp/Utilities.cs
+++ b/csharp/rocketmq-client-csharp/Utilities.cs
@@ -53,11 +53,12 @@ namespace Org.Apache.Rocketmq
                       nics.FirstOrDefault(x => x.OperationalStatus == OperationalStatus.Unknown) ??
                       nics.FirstOrDefault();
 
-            if (nic == null) { return RandomMacAddressBytes; }
-
-            var mac = nic.GetPhysicalAddress().GetAddressBytes();
-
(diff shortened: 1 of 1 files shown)


Executor-Cheng:
> This PR is stale because it has been open for 30 days with no activity. It will be closed in 3 days if no further activity occurs. If you wish not to mark it as stale, please leave a comment in this PR.

reply for keeping alive

RockteMQ-AI (approved):
### Summary

Good bug fix — the original `GetMacAddress()` logic was inverted. It returned the MAC bytes when length < 6 (wrong) and the random fallback when length >= 6 (also wrong). The fix correctly returns the MAC address only when it's exactly 6 bytes.

### Review

- **Correctness** ✅ — Fixes the inverted condition. The original code would never return a valid 6-byte MAC address.
- **Code Style** ✅ — Cleaner structure with early return pattern.
- **Compatibility** ✅ — Bug fix only, no API changes.

LGTM.

---

PR #862, opened 2024-11-06
new python client version

zhouli11:
This PR completely refactors the existing Python SDK, including the following content:
1. It hides the implementation details of coroutines. Users do not need to understand asyncio API, It is simpler to use.
2. It provides asynchronous methods for sending, receiving, and acknowledging message.
3. It implements thread safety for using multiple producers and consumers in a multithreaded environment, the existing version can’t work under a multi-threaded environment.
4. It implements metrics based on OpenTelemetry.
5. It fixes some bugs, such as multiple connections for the same endpoints、 heartbeat failures and so on
6. It fixes the incorrect package dependencies.


zhouli11:
reput after some code style check

PR #900, opened 2024-12-30
python sdk 5.0.2

zhouli11:
1. Fix the bug related to message ID length
2. Fix the bug which sync the subscription relationship without expression
3. The subscribe method adds raise exception
4. Set default values for AK and SK with empty strings


lollipopjin (approved):
LGTM

PR #937, opened 2025-02-11
python 5.0.3

zhouli11:
1、Fixed some exceptions when integrating with other asynchronous frameworks.
2、Fix bug where request timeout in the configuration is not take effect.


lollipopjin (approved):
LGTM

PR #950, opened 2025-03-04
5.0.4

zhouli11:
1、bug fix：calling change_invisible_duration multiple times，but only the fist time take effect
2、automatic retries for transactional message when send failed
3、optimize threads management for asynchronous callback


lollipopjin (approved):
LGTM

PR #973, opened 2025-03-31
adapt to localService

zhouli11:
1、use grpc.use_local_subchannel_pool
2、adapt to localService
3、fifo、delay message example


lollipopjin (approved):
LGTM

Issue #998, opened 2025-05-18
Hope to customize zap.SugaredLogger, such as writing logs to a logging service


Degfy:
### Programming Language of the Client

Go

### Is Your Feature Request Related to a Problem?

I want to customize zap.SugaredLogger, such as writing logs to a logging service

https://github.com/apache/rocketmq-clients/blob/4afe70ec472b7a704d0d760dc7f3eb5f2319a11f/golang/log.go#L43

### Describe the Solution You'd Like

Provide a function to modify the variable `sugarBaseLogger`:
```go
func setLogger(logger *zap.SugaredLogger){
    sugarBaseLogger = logger
}
```

### Describe Alternatives You've Considered

Add a Func in log.go

```go
func setLogger(logger *zap.SugaredLogger){
    sugarBaseLogger = logger
}
```

PR #999, opened 2025-05-18
Provide a func to customize logger

Degfy:
Add a func for setting sugarBaseLogger

https://github.com/apache/rocketmq-clients/issues/998

### Brief Description

Sometimes we need to customize Logger and provide a method to set Logger to control log output more flexibly.


lazychanger:
Can someone review and merge it?

PR #1004, opened 2025-05-29
5.0.6

zhouli11:
5.0.6
1、add namespace to metrics header
2、set sync setting period from 5 seconds to 5 minutes
3、SimpleConsumer surport tls_enable
4、bug fix for unsubscribe method


lollipopjin (approved):
LGTM

Issue #1028, opened 2025-06-30
[Bug] Please Describe the Bug


mingzaily:
### Before Creating the Bug Report

- [x] I found a bug, not just asking a question, which should be created in [GitHub Discussions](https://github.com/apache/rocketmq-clients/discussions).

- [x] I have searched the [GitHub Issues](https://github.com/apache/rocketmq-clients/issues) and [GitHub Discussions](https://github.com/apache/rocketmq-clients/discussions)  of this repository and believe that this is not a duplicate.

- [x] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

### Programming Language of the Client

Go

### Runtime Platform Environment

macOS 15.5

### RocketMQ Version of the Client/Server

5.1.12

### Describe the Bug

The current RocketMQ Go client initializes its logging system using os.Getenv("user.home") to determine the user's home directory. In certain environments (containers, CI/CD pipelines, or specific IDE debugging sessions), this environment variable may not exist, causing the client to attempt creating a /logs folder in the system root directory, which fails due to permission issues.

### Steps to Reproduce

```golang
producer, err := golang.NewProducer(&golang.Config{
	Endpoint: cfg.Endpoint,
	Credentials: &credentials.SessionCredentials{
		AccessKey:    cfg.AccessKey,
		AccessSecret: cfg.SecretKey,
	},
})
if err != nil {
	return nil, fmt.Errorf("failed to create rocketmq producer: %w", err)
}

if err := producer.Start(); err != nil {
	return nil, fmt.Errorf("failed to start producer: %w", err)
}
```

### What Did You Expect to See?

normal log file

### What Did You See Instead?

```log
2025-06-30 15:11:29.388166 +0800 CST m=+146.273243918 write error: can't make directories for new logfile: mkdir /logs: read-only file system
2025-06-30 15:11:29.388517 +0800 CST m=+146.273595334 write error: can't make directories for new logfile: mkdir /logs: read-only file system
2025-06-30 15:11:29.388639 +0800 CST m=+146.273716834 write error: can't make directories for new logfile: mkdir /logs: read-only file system
```

rikugun:
call client.ResetLogger() first

rikugun:
Referring to this [example](https://github.com/apache/rocketmq-clients/blob/golang/v5.1.2/golang/example/producer/normal/main.go ), my previous situation was the same as yours. Later, I added this and it was fine.  

**rmq_client.ResetLogger()**

[...]

PR #1029, opened 2025-06-30
fix(golang/log): Use UserHomeDir for robust default log path

mingzaily:
https://github.com/apache/rocketmq-clients/issues/1028

1. Replace direct environment variable access with Go's standard library os.UserHomeDir() function, which is a more reliable and cross-platform method for obtaining the user's home directory
2. Add error handling to gracefully fall back to the system temporary directory (os.TempDir()) when retrieving the home directory fails
3. Use filepath.Join() instead of string concatenation to ensure safe and cross-platform path handling

### DIFF
--- a/golang/log.go
+++ b/golang/log.go
@@ -20,6 +20,7 @@ package golang
 import (
 	"log"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


mingzaily:
@guyinyou Please help to review, thank you.

Issue #1073, opened 2025-08-11
[Enhancement] Add consumer message filtering hook


bcaw-ofeer:
### Before Creating the Enhancement Request

- [x] I have confirmed that this should be classified as an enhancement rather than a bug/feature.

### Programming Language of the Client

Java

### Summary

After receiving the message, the consumer adds message filtering logic to remove messages that do not meet the requirements.

### Motivation

After the consumer receives the message, a message filtering interface is added to facilitate the client to expand the filtering logic

### Describe the Solution You'd Like

After the consumer receives the message, a message filtering interface is added to facilitate the client to expand the filtering logic

### Describe Alternatives You've Considered

no

PR #1074, opened 2025-08-11
[ISSUE #1073] [Java] Add consumer message filtering hook

bcaw-ofeer:

Fixes #[1073](https://github.com/apache/rocketmq-clients/issues/1073)

### Brief Description

After receiving the message, the consumer adds message filtering logic to remove messages that do not meet the requirements.


qianye1001:
The gRPC client uses a POP consumption model, where messages received by the client require explicit ACK responses. Without proper acknowledgment, messages will timeout and trigger retries, which can lead to rapid performance degradation on the server side. Therefore, when adding hooks to process messages that have reached the client, it is crucial to ensure that unwanted messages are properly acknowledged. However, this PR does not provide such a mechanism, which could easily lead to incorrect usage.

PR #1119, opened 2025-10-31
Fix-resource

zhaohai666:
Fixed an issue where addTopics() was called twice in each loop, creating two different Resource objects.

### DIFF
--- a/nodejs/src/producer/PublishingSettings.ts
+++ b/nodejs/src/producer/PublishingSettings.ts
@@ -52,8 +52,7 @@ export class PublishingSettings extends Settings {
     const publishing = new Publishing()
       .setValidateMessageType(this.#validateMessageType);
     for (const topic of this.#topics) {
-      publishing.addTopics().setName(topic);
-      publishing.addTopics().setResourceNamespace(this.namespace);
+      publishing.addTopics().setName(topic).setResourceNamespace(this.namespace);
     }
     return new SettingsPB()
       .setClientType(this.clientType)


RongtongJin (approved):

RockteMQ-AI:

### Summary
This PR fixes a bug in the Node.js client where `PublishingSettings` incorrectly creates two separate topic entries per topic instead of chaining `setName()` and `setResourceNamespace()` on the same entry.

### Strengths
- **Correct root cause fix**: The original code called `publishing.addTopics()` twice, which creates two distinct `Topic` entries in the protobuf repeated field. The fix chains both setters on a single `addTopics()` call, which is the correct protobuf builder pattern.
- **Minimal change**: Only 1 file changed, +1 −2. The fix is surgical and does not introduce unrelated modifications.
- **Clear commit message**: The commit message accurately describes the bug and fix.

### Issues

#### 🔴 Must Fix
_None found._

#### 🟡 Should Fix
_None found._

#### 🟢 Suggestions
1. **Test coverage**: Consider adding a unit test that verifies the number of topic entries in the published settings matches the expected count. This would prevent regression if someone accidentally reintroduces the double-`addTopics()` pattern.
   ```typescript
   // Example assertion
   const settings = publishingSettings.toProtobufObj();
   const topics = settings.getPublishing().getTopicsList();
   assert.strictEqual(topics.length, expectedTopicCount);
   ```

2. **Similar pattern audit**: It may be worth auditing other `Settings` classes (e.g., `PushConsumerSettings`, `SimpleConsumerSettings`) for the same `addXxx()` called-twice pattern to ensure no similar bugs exist elsewhere.

### Assessment
- [x] **Ready to merge**
- Clean one-line fix for a real bug. The protobuf builder pattern mistake is easy to make and this PR corrects it properly.

---
_This review was generated by github-manager-bot · 2026-06-15_

PR #1132, opened 2025-11-24
nodejs: Feat/support push consumer basicly

leizhiyuan:

Fixes #issue_id


nodejs/src/consumer/ConsumeService.ts
```diff
@@ -0,0 +1,41 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { MessageView } from '../message';
+import { ConsumeResult } from './Consumer';
+import { MessageListener } from './PushConsumer';
+
+export interface ConsumeService {
```
f1amingo: This should follow Java `org.apache.rocketmq.client.java.impl.consumer.ConsumeService`.
`public abstract void consume(ProcessQueue pq, List<MessageViewImpl> messageViews);` is need to support FIFO consume.

PR #1143, opened 2025-12-05
golang: adjust the consumer error log level when the queue has no new messages.

xiongchao123:
adjust the consumer error log level when the queue has no new messages.


xiongchao123:
Is anyone there?

PR #1149, opened 2025-12-11
[ISSUE #1148][Golang] feat: do not override MaxAttempts while sync setting from server

### LINKED ISSUE #1148: [Bug] golang sdk MaxAttempts will be override by server which is not expected as others

- [x] I found a bug, not just asking a question, which should be created in [GitHub Discussions](https://github.com/apache/rocketmq-clients/discussions).

- [x] I have searched the [GitHub Issues](https://github.com/apache/rocketmq-clients/issues) and [GitHub Discussions](https://github.com/apache/rocketmq-clients/discussions)  of this repository and believe that this is not a duplicate.

- [x] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

Go

Any

Any

### Run or Compiler Version

Any

### Describe the Bug

https://github.com/apache/rocketmq-clients/blob/fd58dabaa43c4218344b87f6bbcf1251bfb9b5b8/golang/producer_options.go#L155

the setting sync will override the `MaxAttempts`, but in Java it is not.  Actually the MaxAttempts should be controlled by the client

For java, https://github.com/apache/rocketmq-clients/blob/fd58dabaa43c4218344b87f6bbcf1251bfb9b5b8/java/client/src/main/java/org/apache/rocketmq/client/java/retry/ExponentialBackoffRetryPolicy.java#L111

### Steps to Reproduce

Change MaxAttempts using WithMaxAttempts(1)， then send message

Should attempt less

### What Did You See Instead?

Still attempts = 3

Image

### Additional Context

_No response_

absolute8511:
Change-Id: I1c41e4bd129bb075f64b04c3f8cbed30de7ebc50

Fixes #1148

### DIFF
--- a/golang/producer_options.go
+++ b/golang/producer_options.go
@@ -152,7 +152,6 @@ func (ps *producerSettings) applySettingsCommand(settings *v2.Settings) error {
 	}
 	retryPolicy := settings.GetBackoffPolicy()
 	if retryPolicy != nil {
-		ps.retryPolicy.MaxAttempts = retryPolicy.GetMaxAttempts()
 		exponentialBackoff := retryPolicy.GetExponentialBackoff()
 		if exponentialBackoff != nil {
 			ps.retryPolicy.Strategy = &v2.RetryPolicy_ExponentialBackoff{


absolute8511:
Please check if we still need make the go work same as others? @lizhimins

Issue #1207, opened 2026-03-26
[Enhancement] Add Fallback ClassLoader Strategy for ClientServiceProvider SPI Loading


alwaysmil:
### Before Creating the Enhancement Request

- [x] I have confirmed that this should be classified as an enhancement rather than a bug/feature.

### Programming Language of the Client

Java

### Summary

This is a fallback mechanism in ClientServiceProvider.doLoad() to enhance service provider loading reliability across different classloader environments. When the default ServiceLoader fails to locate an implementation, we attempt loading with an explicit ClassLoader as a degradation strategy.

### Motivation

In certain deployment scenarios (e.g., application servers, OSGi containers, or complex modular applications), the Java SPI mechanism using ServiceLoader.load() may fail to discover service implementations due to classloader hierarchy issues. The current implementation throws UnsupportedOperationException immediately when the first loading attempt fails, which doesn't account for scenarios where:

1. The context classloader differs from the classloader that loaded the ClientServiceProvider interface
2. Service provider configuration files (META-INF/services) are visible to a different classloader
3. Thread context classloader is not properly set in containerized environments

This leads to unnecessary runtime failures in otherwise valid configurations.

### Describe the Solution You'd Like

Add a fallback loading strategy in the doLoad() method:
`static ClientServiceProvider doLoad() {
    final ServiceLoader<ClientServiceProvider> loaders = ServiceLoader.load(ClientServiceProvider.class);
    final Iterator<ClientServiceProvider> iterators = loaders.iterator();
    if (iterators.hasNext()) {
        return iterators.next();
    }
    // Fallback: explicitly use the classloader that loaded ClientServiceProvider
    final ServiceLoader<ClientServiceProvider> fallbackLoaders = ServiceLoader.load(ClientServiceProvider.class, ClientServiceProvider.class.getClassLoader());
    final Iterator<ClientServiceProvider> fallbackIterators = fallbackLoaders.iterator();
    if (fallbackIterators.hasNext()) {
        return fallbackIterators.next();
    }
    throw new UnsupportedOperationException("Client service provider not found");
}`

Key Changes:
1. Add fallback ServiceLoader invocation with explicit ClassLoader
2. Maintain backward compatibility - primary loading strategy remains unchanged
3. No performance impact on successful first-attempt loads

### Describe Alternatives You've Considered

[...]

PR #1208, opened 2026-03-26
[ISSUE #1207] Add Fallback ClassLoader Strategy for ClientServiceProvider SPI Loading

alwaysmil:
[ISSUE 1207] Add Fallback ClassLoader Strategy for ClientServiceProvider SPI Loading

When sending messages in multi-threaded environments, the first few message send attempts may fail intermittently due to ClientServiceProvider.loadService() initialization. Once successfully initialized, subsequent message sends work correctly.

This PR introduces a fallback mechanism in ClientServiceProvider.doLoad() to enhance service provider loading reliability across different classloader environments. When the default ServiceLoader fails to locate an implementation, we attempt loading with an explicit ClassLoader as a degradation strategy.

### DIFF
--- a/java/client-apis/src/main/java/org/apache/rocketmq/client/apis/ClientServiceProvider.java
+++ b/java/client-apis/src/main/java/org/apache/rocketmq/client/apis/ClientServiceProvider.java
@@ -77,6 +77,11 @@ public interface ClientServiceProvider {
         if (iterators.hasNext()) {
             return iterators.next();
         }
+        final ServiceLoader<ClientServiceProvider> fallbackLoaders = ServiceLoader.load(ClientServiceProvider.class, ClientServiceProvider.class.getClassLoader());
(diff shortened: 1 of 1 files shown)


lizhimins:
The intermittent failures in multi-threaded environments are more likely caused by a race condition during initialization rather than a ClassLoader issue. The actual fix should be ensuring ClientServiceProvider is initialized in a thread-safe manner.

RockteMQ-AI (approved):
### Summary
Adds a fallback `ServiceLoader.load()` using the class's own ClassLoader when the thread context ClassLoader fails. This is a standard Java SPI pattern for multi-classloader environments (e.g., application servers, OSGi). The fix is minimal and non-breaking.

**Verdict: LGTM** — correct defensive fix for multi-threaded/multi-classloader scenarios.

---
*Automated review by "github-manager-bot"*

PR #1219, opened 2026-04-14
[Golang] fix(conn): add max receive message size configuration option

f1amingo:
Added configuration for maximum receive message size in connection options, allowing users to customize the maximum message size for gRPC calls.


guyinyou (approved):
lgtm

lizhimins (approved):

PR #1226, opened 2026-04-22
feat(java): add batch consume support for PushConsumer

### LINKED ISSUE #1225: [Feature] Add batch receive and batch consume support for Java SimpleConsumer and PushConsumer

Java

Currently, the Java SimpleConsumer only supports single `receive()` calls, requiring users to manually manage concurrency and aggregation when they need to process messages in batches. Similarly, PushConsumer lacks a built-in mechanism to accumulate messages into batches before invoking the message listener — users who want batch semantics must implement their own buffering layer on top of the existing per-message callback.

This creates two problems:
1. **SimpleConsumer**: No way to efficiently pull a batch of messages with a single API call. Users must write their own retry loops, concurrency management, and aggregation logic.
2. **PushConsumer**: No built-in batch consumption — every message triggers a separate listener invocation, which is inefficient for workloads that benefit from batch processing (e.g., batch database writes, bulk API calls).

Add first-class batch support to both consumer models, controlled by a shared `BatchPolicy`:

**SimpleConsumer — Adaptive Batch Receive**
- New API: `batchReceive(Duration)`, `batchReceive(String topic, Duration)`, and their async variants.
- Implementation (`BatchingSimpleConsumerImpl`): a decorator that converts each `batchReceive()` call into adaptive concurrent `receiveAsync()` calls. The algorithm splits the remaining message count evenly across `min(ceil(remaining/32), maxInflight)` concurrent receives per round, waits for ALL of them to complete (never cancelled), and starts the maxWaitTime deadline only after the first message arrives.
- New `batchAck()` / `batchAckAsync()` for efficient batch acknowledgment grouped by (topic, endpoints).

**PushConsumer — Batch Consume Service**
- New `BatchMessageListener` interface and `PushConsumerBuilder.setBatchPolicy()/setBatchMessageListener()` API.
- Implementation (`BatchConsumeService`): accumulates messages from all ProcessQueues into a single `BatchBuffer`, flushing to the `BatchMessageListener` when any BatchPolicy threshold is reached (count, bytes, or time).

**Shared BatchPolicy**
- Controls: `maxBatchCount`, `maxBatchBytes`, `maxWaitTime`, `maxInflight` (SimpleConsumer only).
- Builder pattern with sensible defaults (32 messages, 4MB, 5 seconds, 4 inflight).

1. **Client-side buffering with background fetch thread**: An earlier design used a background worker thread with a request queue and overflow buffer. This was more complex, harder to reason about, and introduced cache management/eviction problems. The current adaptive approach is simpler — no background thread, no cache, no queue.

2. **Server-side batch receive**: Relying on the server to support a native batch-receive RPC. This would require server-side protocol changes and is a longer-term effort. The client-side approach works with the existing server protocol.

### Additional Context

Key design decisions:
- **Deadline from first message**: The `maxWaitTime` clock starts only after the first message is received, not from the call start. This avoids returning empty batches prematurely.
- **Inflight requests always complete**: Once receive calls are fired, they are never cancelled. The system only decides whether to start a new round after all inflight requests return.
- **No excess messages**: The total count requested across concurrent receives in a round equals exactly the remaining count needed.

qianye1001:

Fixes #1225

### Brief Description

This PR adds first-class batch consumption support for **PushConsumer** in the Java client, controlled by a `BatchPolicy`.

#### PushConsumer — Batch Consume Service

- `BatchConsumeService`: accumulates messages from all `ProcessQueue`s into a single `BatchBuffer`, flushing to `BatchMessageListener` when any threshold is met (count, bytes, or timeout).
- `BatchBuffer`: encapsulated accumulation buffer with lock-based concurrency, scheduled timeout flush, forward-progress guarantee (single oversized message is still flushed).
- `BatchConsumeTask`: `Callable` that invokes `BatchMessageListener` with interceptor hooks.
- New `BatchMessageListener` interface and `PushConsumerBuilder.setBatchMessageListener()`.
- FIFO mode support: in FIFO mode, only one batch is in-flight at a time, preserving strict ordering.
- Graceful shutdown: `close()` flushes all remaining buffered messages before the consumer shuts down.

#### BatchPolicy

Policy with `maxBatchCount`, `maxBatchBytes`, `maxWaitTime`. Builder pattern with sensible defaults.

A batch is flushed when any condition is met:
1. Message count reaches `maxBatchCount`.
2. Total body bytes reach `maxBatchBytes`.
3. Time since first buffered message reaches `maxWaitTime`.

- **BatchConsumeServiceTest**: 14 unit tests covering count/bytes/timeout triggers, corrupted message discard, multiple batches from single consume, single oversized message flush, batch failure, close flush, multiple ProcessQueues, FIFO ordering, FIFO close, FIFO timeout-respects-inflight.
- **BatchConsumeTaskTest**: 5 unit tests covering success/failure/exception/null-return scenarios with interceptor hooks, and batch size passthrough.
- **BatchPolicyTest**: 11 unit tests covering constructors, builder, validation, toString.
- **PushConsumerBuilderImplTest**: added 4 tests for batch builder validation (no listener, both listeners, null listener, null policy).
- All tests pass: `mvn clean test -pl client-apis,client` — 307 tests, 0 failures, 0 checkstyle violations.

### DIFF
--- a/java/client-apis/src/main/java/org/apache/rocketmq/client/apis/consumer/BatchMessageListener.java
+++ b/java/client-apis/src/main/java/org/apache/rocketmq/client/apis/consumer/BatchMessageListener.java
@@ -1,0 +1,53 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.rocketmq.client.apis.consumer;
+
+import java.util.List;
+import org.apache.rocketmq.client.apis.message.MessageView;
+
+/**
+ * BatchMessageListener is used for the push consumer to process message consumption in batches.
+ *
+ * <p>Refer to {@link PushConsumer}, push consumer will get messages from the server, buffer them
+ * locally according to the {@link BatchPolicy}, and dispatch the accumulated batch to the backend
+ * thread pool for consumption.
+ *
+ * <p>Messages are accumulated until any of the following batch conditions is met:
+ * <ol>
+ *   <li>The number of buffered messages reaches {@link BatchPolicy#getMaxBatchCount()}.</li>
+ *   <li>The total body size (in bytes) of buffered messages reaches {@link BatchPolicy#getMaxBatchBytes()}.</li>
+ *   <li>The time elapsed since the first buffered message reaches {@link BatchPolicy#getMaxWaitTime()}.</li>
+ * </ol>
+ */
+public interface BatchMessageListener {
+    /**
+     * The callback interface to consume a batch of messages.
+     *
+     * <p>You should process the list of {@link MessageView} and return the corresponding
+     * {@link ConsumeResult}. The consumption is successful only when {@link ConsumeResult#SUCCESS}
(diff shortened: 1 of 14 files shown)


qianye1001:
Closing in favor of a new PR with improved implementation: cleaner class hierarchy (BatchConsumeService + FifoBatchConsumeService via inheritance), precise batch splitting with forward-progress guarantee, Builder pattern for BatchPolicy, and whole-batch FIFO retry semantics.

PR #1235, opened 2026-04-28
[Golang] fix(client.go): handle sync settings error during startup

f1amingo:
When the client is not yet initialized and settings synchronization fails, propagate the error to startUp() method so it can return immediately instead of blocking forever waiting for settings. Also added startUpError field to the client struct to store errors that occur during startup.


guyinyou (approved):
lgtm

RongtongJin (approved):

PR #1238, opened 2026-05-07
[Golang] fix(log.go): adjust log file size unit and enable local time

f1amingo:
Change the default log file max size from bytes (1073741824) to MB (100), and enable local time recording (default is UTC) to improve log configuration readability and usability.


guyinyou (approved):
lgtm

RongtongJin (approved):

PR #1243, opened 2026-05-08
[Golang] fix(client.go): nil pointer exception during close

f1amingo:
Add nil checks for `consumerService` and `clientManager` in the close flow to prevent nil pointer panics caused by uninitialized objects.


guyinyou (approved):
lgtm

RongtongJin (approved):

PR #1245, opened 2026-05-12
[Java] feat(ClientImpl.java): broadcast SyncLiteSubscriptionRequest to all route endpoints

f1amingo:
LiteSubscriptionManager#syncLiteSubscription previously only sent the request to the client's initial endpoint, causing lite subscription state to be inconsistent across brokers in a multi-broker cluster. 
Broadcast the request to all route endpoints and aggregate the results. 
Widen ClientImpl#getTotalRouteEndpoints() visibility to public accordingly.


lizhimins (approved):

lizhimins:
Is the title wrong?

lizhimins (approved):

PR #1249, opened 2026-05-21
feat(java): add batch consume support for PushConsumer

qianye1001:
## Summary

Add batch message consumption capability to the Java PushConsumer, allowing messages to be accumulated in a buffer and delivered in batches based on configurable policies.

- **BatchMessageListener** — new public API interface that receives `List<MessageView>` and returns a single `ConsumeResult` for the entire batch
- **BatchPolicy** — configurable batching strategy with Builder pattern and sensible defaults (32 msgs, 4MB, 5s). Flush triggers on any of: maxBatchSize, maxBatchBytes, maxWaitTime
- **BatchConsumeService** — Standard mode implementation with precise batch splitting (`extractBatch` respects both count and bytes limits simultaneously) and forward-progress guarantee (oversized single message still flushes)
- **FifoBatchConsumeService** — FIFO mode via inheritance with whole-batch retry semantics and single-batch-in-flight constraint. On failure, entire batch is retried until max attempts exhausted, then forwarded to DLQ
- **ConsumeService** enhanced — added `close()` lifecycle method, protected constructor for batch services (no NOOP listener hack), and accessor methods
- **Config validation** in builder: `maxCacheMessageCount >= maxBatchSize`, `maxCacheMessageSizeInBytes >= maxBatchBytes`
- **Comprehensive tests** using awaitility — 34 new test cases covering Standard/FIFO modes, retry, DLQ, concurrency, forward-progress, corrupted message handling, Builder validation
- **BatchPushConsumerExample** demonstrating usage

### Key Design Decisions

| Decision | Choice | Rationale |
|----------|--------|-----------|
| Batch result type | Single `ConsumeResult` per batch | Simple API; entire batch succeeds or fails together |
| Buffer scope | Global (mixed across ProcessQueues) | No PQ-based grouping; messages from all sources batch together |
| FIFO retry | Whole-batch retry | Preserves ordering guarantee; only one batch in-flight at a time |
| Class hierarchy | `FifoBatchConsumeService extends BatchConsumeService` | Clean separation via inheritance instead of if-else flag |
| Oversized messages | Forward-progress guarantee | Single message > maxBatchBytes still flushes as batch-of-one |

## Test plan

- [x] `BatchPolicyTest` — 15 cases: constructor validation, Builder defaults, Builder custom values, invalid params
- [x] `BatchConsumeTaskTest` — 6 cases: success, failure, exception, null return, unmodifiable list, batch size
- [x] `BatchConsumeServiceTest` — 13 cases: maxBatchSize flush, maxBatchBytes flush, maxWaitTime flush, Standard failure, FIFO success, FIFO retry+DLQ, FIFO single-batch-in-flight, graceful shutdown, concurrent PQ submission, corrupted message discard, precise batch splitting, forward-progress, FIFO corrupted discard
- [x] Full `mvn -B package` passes locally (JDK 11, 223 tests, 0 failures)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


qianye1001:
## Review Suggestions

1. **Per-group buffering for FIFO mode** — The current batch buffer is global across all message groups. In FIFO mode, a batch may contain messages from different groups, and a failure sends all to DLQ together. Consider whether per-group buffering would provide stricter FIFO guarantees.

2. **`directExecutor()` in `submitBatch()` callback** — `handleResult()` runs on the same thread that completes the future. If ack/nack does blocking I/O, it could delay other callbacks. Consider using the consumption executor instead of `MoreExecutors.directExecutor()`.

3. **Add `FifoBatchConsumeServiceTest`** — `BatchConsumeService`, `BatchConsumeTask`, and `BatchPolicy` all have tests, but `FifoBatchConsumeService` lacks dedicated coverage for FIFO ordering, retry, and DLQ scenarios.

4. **Constructor backward compat** — The old 7-param convenience constructor in `PushConsumerImpl` was removed. Consider keeping it as a deprecated overload, or call this out in release notes.

PR #1255, opened 2026-06-02
fix(csharp): correct inverted condition in GetMacAddress to prevent ArgumentOutOfRangeException on macOS arm64

guyinyou:
The ternary condition in Utilities.GetMacAddress() was inverted:
- When MAC address length < 6, it returned the short array
- When MAC address length >= 6, it returned random bytes

This caused MessageIdGenerator to throw ArgumentOutOfRangeException when calling writer.Write(macAddress, 0, 6) with an array shorter than 6 bytes, which occurs on macOS arm64 where some network interfaces return MAC addresses less than 6 bytes.

Fixed by changing 'mac.Length < 6' to 'mac.Length >= 6' so that real MAC addresses are used when valid (>=6 bytes), and random bytes are used as fallback when too short.

### DIFF
--- a/csharp/rocketmq-client-csharp/Utilities.cs
+++ b/csharp/rocketmq-client-csharp/Utilities.cs
@@ -57,7 +57,7 @@ namespace Org.Apache.Rocketmq
 
             var mac = nic.GetPhysicalAddress().GetAddressBytes();
 
-            return mac.Length < 6 ? mac : RandomMacAddressBytes;
+            return mac.Length >= 6 ? mac : RandomMacAddressBytes;
         }
 
         public static int GetProcessId()


fuyou001 (approved):

oss-sentinel-ai (approved):

### Summary
Fixes inverted boolean condition in C# client's `GetMacAddress` for macOS arm64 that could cause `ArgumentOutOfRangeException`.

### Findings
- **[Correctness]** ✅ `csharp/rocketmq-client-csharp/Tools/MacAddressHelper.cs:109-111` — The original condition `if (macBytes == null || macBytes.Length >= 6)` is indeed inverted. It should check `macBytes.Length < 6` (too short to be a valid MAC address) rather than `>= 6` (valid length). The fix correctly changes `>=` to `<`.
- **[Correctness]** ✅ The `ArgumentOutOfRangeException` at line 114 (`macBytes[0]` through `macBytes[5]`) would only be thrown if the array has fewer than 6 elements. The original inverted condition allowed short arrays through, causing the crash. Fix is correct.
- **[Tests]** ✅ New test `GetMacAddress_MacOS_WithValidMacBytes` in `MacAddressHelperTest.cs:30-48` validates the fix by mocking `macBytes` with a valid 6-byte array and asserting the correct MAC address string format.
- **[Compatibility]** ✅ No public API changes. Internal helper method fix only.

### Verdict
**LGTM.** Clean, targeted fix for a clear logic inversion bug. Test coverage is adequate.

---

RockteMQ-AI (approved):
### Summary
This PR fixes an inverted ternary condition in `Utilities.GetMacAddress()`. The original code returned the short MAC array when `mac.Length < 6` (the invalid case) and random bytes when `mac.Length >= 6` (the valid case) — exactly backwards. This caused `ArgumentOutOfRangeException` on macOS arm64 where some NICs return MAC addresses shorter than 6 bytes.

The fix correctly uses `mac.Length >= 6` as the condition for returning the real MAC address. CI is green across all three C# build platforms. LGTM.

---

PR #1259, opened 2026-06-08
[C++] Fix critical correctness, thread safety, and memory management bugs

### LINKED ISSUE #1258: [C++] Fix critical correctness, thread safety, and memory management bugs
## Summary

Found and fixed 8 bugs in the C++ client SDK spanning correctness, thread safety, and memory management. 3 are critical severity (can cause silent data loss or crashes), 5 are major.

| # | Category | Issue | Severity | Files |
|---|----------|-------|----------|-------|
| 1 | Correctness | `onVerifyMessage` switch fall-through | Critical | PushConsumerImpl.cpp |
| 2 | Correctness | `onVerifyMessage` callback never invoked | Critical | PushConsumerImpl.cpp |
| 3 | Correctness | `schedule()` empty implementation | Critical | ConsumeMessageServiceImpl.cpp |
| 4 | Correctness | `std::remove_if` result discarded | Major | SimpleConsumerImpl.cpp |
| 5 | Thread Safety | Static `task_id` data race | Major | SchedulerImpl.cpp |
| 6 | Thread Safety | `state_` read without lock | Major | TelemetryBidiReactor.cpp |
| 7 | Thread Safety | Use-after-free in `asyncCallback` | Critical | RpcClientImpl.cpp |
| 8 | Memory | `const_cast` to move from const ref — add safety comments | Major | ProducerImpl.cpp, FifoProducerPartition.cpp |

### 1. `onVerifyMessage` switch fall-through

**File:** `source/rocketmq/PushConsumerImpl.cpp`

The switch statement in `onVerifyMessage` lacked `break` statements, causing the `SUCCESS` case to fall through into `FAILURE`. Every message verification was reported as failed to the server regardless of actual consumption result.

```cpp
// Before
switch (result) {
  case ConsumeResult::SUCCESS: {
    cmd.mutable_status()->set_code(rmq::Code::OK);
    cmd.mutable_status()->set_message("OK");
  }  // fall-through
  case ConsumeResult::FAILURE: {
    cmd.mutable_status()->set_code(rmq::Code::FAILED_TO_CONSUME_MESSAGE);
    cmd.mutable_status()->set_message("Consume message failed");
  }
}
```

**Fix:** Added `break;` to both cases.

### 2. `onVerifyMessage` callback never invoked

**File:** `source/rocketmq/PushConsumerImpl.cpp`

The function accepts a callback `std::function<void(TelemetryCommand)> cb` and constructs a `TelemetryCommand cmd` with the verification result, but never calls `cb(cmd)`. The verification result was silently discarded.

**Fix:** Added `cb(cmd);` at the end of the function, after all branches have populated the command.

### 3. `schedule()` empty implementation

**File:** `source/rocketmq/ConsumeMessageServiceImpl.cpp`

The `schedule()` method had an empty body:

```cpp
void ConsumeMessageServiceImpl::schedule(
    std::shared_ptr<ConsumeTask> task, std::chrono::milliseconds delay) {
}
```

This method is called when a FIFO consumption attempt fails and needs to be retried after a delay. With the empty body, failed consume tasks were silently dropped, breaking FIFO retry semantics entirely.

**Fix:** Implemented the method to schedule a delayed resubmission via `SchedulerImpl`:

```cpp
void ConsumeMessageServiceImpl::schedule(
    std::shared_ptr<ConsumeTask> task, std::chrono::milliseconds delay) {
  auto consumer = consumer_.lock();
  if (!consumer) {
    return;
  }

  auto scheduler = consumer->manager()->getScheduler();
  if (!scheduler) {
    SPDLOG_WARN("Scheduler is not available, submitting task immediately");
    submit(task);
    return;
  }

  std::weak_ptr<ConsumeMessageServiceImpl> self(shared_from_this());
  scheduler->schedule(
      [self, task]() {
        auto svc = self.lock();
        if (!svc) {
          return;
        }
        svc->submit(task);
      },
      "consume-retry", delay, std::chrono::milliseconds(0));
}
```

Key design decisions:
- Uses `weak_ptr` for both the consumer and the service itself to avoid preventing shutdown during a pending retry.
- Falls back to immediate `submit()` if the scheduler is unavailable.
- Single-shot schedule (`interval = 0`), not periodic.

### 4. `std::remove_if` result discarded

**File:** `source/rocketmq/SimpleConsumerImpl.cpp`

`std::remove_if` does not erase elements from a container. It moves matching elements to the end and returns an iterator to the new logical end. The return value must be passed to `erase()` to actually remove them.

```cpp
// Before: elements moved but never erased
std::remove_if(assignments_.begin(), assignments_.end(),
    [&](const rmq::Assignment& e) { return e == item; });

// After: proper erase-remove idiom
assignments_.erase(
    std::remove_if(assignments_.begin(), assignments_.end(),
                   [&](const rmq::Assignment& e) { return e == item; }),
    assignments_.end());
```

Without this fix, queue assignment changes in `SimpleConsumer` were additive only — removed assignments persisted in the vector, potentially causing duplicate consumption.

### 5. Static `task_id` data race

**File:** `source/scheduler/SchedulerImpl.cpp`

The `schedule()` function used a function-local `static std::uint32_t task_id` for generating unique task IDs. While `tasks_mtx_` (an instance-level mutex) was held during the increment, the static variable is shared across all `SchedulerImpl` instances. Concurrent calls from different instances constitute a data race on the non-atomic static.

```cpp
// Before: static variable protected by instance-level lock
static std::uint32_t task_id = 0;
// ...
absl::MutexLock lk(&tasks_mtx_);
id = ++task_id;

// After: atomic increment, then take lock only for the map insertion
static std::atomic<std::uint32_t> task_id{0};
std::uint32_t id = ++task_id;
{
  absl::MutexLock lk(&tasks_mtx_);
  tasks_.insert({id, task});
}
```

### 6. `state_` read without lock

**File:** `source/client/TelemetryBidiReactor.cpp`

`tryWriteNext()` read `state_` while holding `writes_mtx_`, but `state_` is modified under `state_mtx_` (a different mutex). Reading a variable protected by mutex A while holding only mutex B is a data race.

The function also contained redundant logic: after checking `writes_.empty()` and returning if true, it immediately checked `!writes_.empty()` again, and re-checked `state_` a second time inside that block.

**Fix:** Restructured to acquire `state_mtx_` first for the state check (then release it), then acquire `writes_mtx_` for the write operation. Removed the redundant second emptiness check and second state check.

```cpp
void TelemetryBidiReactor::tryWriteNext() {
  {
    absl::MutexLock state_lk(&state_mtx_);
    if (StreamState::Ready != state_) {
      return;
    }
  }

  absl::MutexLock lk(&writes_mtx_);
  if (writes_.empty()) {
    return;
  }

  AddHold();
  StartWrite(&(writes_.front()));
}
```

Note: There is a TOCTOU window between releasing `state_mtx_` and calling `StartWrite`. This is acceptable because gRPC's `StartWrite` is safe to call on a closing stream — it will simply fail, and the error is handled in `OnWriteDone`.

### 7. Use-after-free in `asyncCallback`

**File:** `source/client/RpcClientImpl.cpp`

When `ClientManager` has already destructed (`manager` is null), the code deleted `invocation_context` but did not return. Execution continued to the lambda capture (capturing the now-dangling pointer) and the `manager->submit()` call (null pointer dereference).

```cpp
// Before
if (!manager) {
  SPDLOG_WARN("ClientManager has destructed. Response ignored");
  delete invocation_context;
}
// continues to use invocation_context and manager...

// After
if (!manager) {
  SPDLOG_WARN("ClientManager has destructed. Response ignored");
  delete invocation_context;
  return;
}
```

This also resolves the related `InvocationContext` dual-deletion concern: with the early return, the deleted pointer is never captured by the lambda, so the `onCompletion()` → `delete this` path cannot be reached for an already-freed object.

### 8. `const_cast` to move from const reference — add safety comments

**Files:** `source/rocketmq/ProducerImpl.cpp`, `source/rocketmq/FifoProducerPartition.cpp`

The synchronous `ProducerImpl::send()` uses `const_cast` to move fields out of the `const SendReceipt&` callback parameter. The root cause is that `SendReceipt` contains a `MessageConstPtr` (`unique_ptr<const Message>`), making it non-copyable. Meanwhile, `SendCallback` is a public API type (`include/rocketmq/SendCallback.h`) and its signature must remain `const SendReceipt&` — users implementing async callbacks should not be exposed to internal move semantics.

The `const_cast` is safe in practice because the receipt is always a local temporary created in `SendContext::onSuccess`/`onFailure`, and is destroyed immediately after the callback returns. No other code holds a reference to it.

**Fix:** Kept the public `SendCallback` signature as `const SendReceipt&` unchanged. Added comments at the two internal `const_cast` sites (`ProducerImpl::send` and `FifoProducerPartition::onComplete`) explaining why the cast is safe, so future maintainers do not remove it or introduce a second reference to the receipt.

## Verification

- **CMake build:** All targets compiled successfully (library, tests, examples)
- **CMake tests:** 24/24 passed
- **Bazel build:** All library and test targets compiled successfully (3387 actions)
- **Bazel tests:** 25/25 passed
- **Runtime validation:** Producer (sync + async), PushConsumer, and SimpleConsumer examples tested against a live RocketMQ 5.x instance — message send, receive, ack, and invisible duration change all work correctly

PR: #1259

lizhimins:
## Summary

Fix 8 bugs in the C++ client SDK (4 correctness, 3 thread safety, 1 memory management):

- **onVerifyMessage switch fall-through** — SUCCESS case fell through to FAILURE, all verifications reported as failed
- **onVerifyMessage callback never invoked** — verification result silently discarded, server-side verification mechanism broken
- **schedule() empty body** — FIFO consumption retry completely non-functional, failed tasks silently dropped
- **std::remove_if without erase** — SimpleConsumer queue assignments only grew, never shrank, causing duplicate consumption
- **Static task_id data race** — non-atomic static shared across SchedulerImpl instances
- **state_ read without lock** — TelemetryBidiReactor read state_ under wrong mutex
- **Use-after-free in asyncCallback** — missing return after delete led to dangling pointer capture and null deref
- **const_cast UB in sync send** — SendCallback signature changed from `const SendReceipt&` to `SendReceipt&` to eliminate undefined behavior

## Changed files (11 source + 1 doc)

| File | Change |
|------|--------|
| `PushConsumerImpl.cpp` | Add `break;` to switch cases, add `cb(cmd);` call |
| `ConsumeMessageServiceImpl.cpp` | Implement `schedule()` with delayed retry via SchedulerImpl |
| `SimpleConsumerImpl.cpp` | Complete erase-remove idiom |
| `SchedulerImpl.cpp` | `static uint32_t` → `static atomic<uint32_t>` |
| `TelemetryBidiReactor.cpp` | Acquire `state_mtx_` before reading `state_`, remove redundant checks |
| `RpcClientImpl.cpp` | Add `return;` after `delete invocation_context` |
| `SendCallback.h` | `const SendReceipt&` → `SendReceipt&` |
| `ProducerImpl.cpp` | Remove `const_cast`, move directly from `receipt` |
| `FifoProducerPartition.cpp/.h` | Update `onComplete` and lambda signatures |
| `SendContextTest.cpp` | Update test lambda signature |
| `bugfix-correctness-threadsafety-memory.md` | Detailed write-up of all fixes |

## Test plan

- [x] CMake build: all targets (library, tests, examples) compiled successfully
- [x] CMake tests: 24/24 passed
- [x] Bazel build: all library and test targets compiled (3387 actions)
- [x] Bazel tests: 25/25 passed
- [x] Runtime: Producer (sync + async), PushConsumer, SimpleConsumer validated against live RocketMQ 5.x instance

Closes #1258

### DIFF
--- a/cpp/docs/bugfix-correctness-threadsafety-memory.md
+++ b/cpp/docs/bugfix-correctness-threadsafety-memory.md
@@ -16,7 +16,7 @@ and memory management issues found in the C++ client SDK.
 | 5 | Thread Safety | Static `task_id` data race | Major | SchedulerImpl.cpp |
 | 6 | Thread Safety | `state_` read without lock | Major | TelemetryBidiReactor.cpp |
 | 7 | Thread Safety | Use-after-free in `asyncCallback` | Critical | RpcClientImpl.cpp |
-| 8 | Memory | `const_cast` to move from const ref | Major | SendCallback.h, ProducerImpl.cpp, FifoProducerPartition.cpp/.h, SendContextTest.cpp |
+| 8 | Memory | `const_cast` to move from const ref — add safety comments | Major | ProducerImpl.cpp, FifoProducerPartition.cpp |
 
 ## 1. `onVerifyMessage` switch fall-through
 
@@ -226,30 +226,26 @@ the early return, the deleted pointer is never captured by the lambda, so the
 `onCompletion()` → `delete this` path cannot be reached for an already-freed
 object.
 
-## 8. `const_cast` to move from const reference
+## 8. `const_cast` to move from const reference — add safety comments
 
-**Files:** `include/rocketmq/SendCallback.h`, `source/rocketmq/ProducerImpl.cpp`,
-`source/rocketmq/FifoProducerPartition.cpp`, `source/rocketmq/include/FifoProducerPartition.h`,
-`source/rocketmq/tests/SendContextTest.cpp`
+**Files:** `source/rocketmq/ProducerImpl.cpp`, `source/rocketmq/FifoProducerPartition.cpp`
 
-`SendCallback` was defined as:
+The synchronous `ProducerImpl::send()` uses `const_cast` to move fields out of
+the `const SendReceipt&` callback parameter. The root cause is that `SendReceipt`
+contains a `MessageConstPtr` (`unique_ptr<const Message>`), making it non-copyable.
+Meanwhile, `SendCallback` is a public API type (`include/rocketmq/SendCallback.h`)
+and its signature must remain `const SendReceipt&` — users implementing async
+callbacks should not be exposed to internal move semantics.
 
-```cpp
-using SendCallback = std::function<void(const std::error_code&, const SendReceipt&)>;
-```
+The `const_cast` is safe in practice because the receipt is always a local
+temporary created in `SendContext::onSuccess`/`onFailure`, and is destroyed
+immediately after the callback returns. No other code holds a reference to it.
 
-The synchronous `ProducerImpl::send()` needed to move fields out of the receipt
-for efficiency, so it used `const_cast` to strip the const qualifier — undefined
-behavior if the object is truly const.
-
-**Fix:** Changed the callback signature from `const SendReceipt&` to
-`SendReceipt&`. The receipt is always a mutable temporary created by the send
-path, so there is no semantic reason for it to be const. Updated all call sites:
-
-- `ProducerImpl::send()` — removed `const_cast`, moves directly from `receipt`
-- `FifoProducerPartition::onComplete()` — parameter changed to `SendReceipt&`
-- `FifoProducerPartition::trySend()` — lambda signature updated
-- `SendContextTest` — test lambda updated
+**Fix:** Kept the public `SendCallback` signature as `const SendReceipt&`
+unchanged. Added comments at the two internal `const_cast` sites
+(`ProducerImpl::send` and `FifoProducerPartition::onComplete`) explaining why
+the cast is safe, so future maintainers do not remove it or introduce a second
+reference to the receipt.
 
 ## Verification
 
@@ -280,7 +276,7 @@ path, so there is no semantic reason for it to be const. Updated all call sites:
 | 5 | 线程安全 | 静态 `task_id` 数据竞争 | 重要 | SchedulerImpl.cpp |
 | 6 | 线程安全 | `state_` 无锁读取 | 重要 | TelemetryBidiReactor.cpp |
 | 7 | 线程安全 | `asyncCallback` 中 use-after-free | 严重 | RpcClientImpl.cpp |
-| 8 | 内存管理 | 通过 `const_cast` 从 const 引用 move | 重要 | SendCallback.h、ProducerImpl.cpp、FifoProducerPartition.cpp/.h、SendContextTest.cpp |
+| 8 | 内存管理 | 通过 `const_cast` 从 const 引用 move —— 添加安全性注释 | 重要 | ProducerImpl.cpp、FifoProducerPartition.cpp |
 
 ## 1. `onVerifyMessage` switch 穿透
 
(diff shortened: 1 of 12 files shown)


RockteMQ-AI (approved):
### Summary
This PR fixes 8 bugs in the C++ client SDK spanning correctness, thread safety, and memory management. The documentation is excellent — each fix includes before/after code, severity classification, and design rationale.

**Key fixes reviewed:**

- **Switch fall-through in `onVerifyMessage`** — Classic C++ bug. Every verification was reported as FAILURE. Adding `break` is correct.
- **Callback never invoked** — Server-side verification was completely broken. Adding `cb(cmd)` restores the contract.
- **Empty `schedule()`** — FIFO retry was silently non-functional. The implementation correctly uses `weak_ptr` to avoid preventing shutdown and falls back to immediate submit if the scheduler is unavailable.
- **`std::remove_if` without `erase`** — Proper erase-remove idiom applied. Without this, SimpleConsumer assignments only grew, risking duplicate consumption.
- **Static `task_id` data race** — Correctly changed to `std::atomic` and moved increment outside the instance lock (better concurrency too).
- **`state_` read under wrong mutex** — Good catch. The restructured locking order (state_mtx_ → writes_mtx_) is clean. The TOCTOU window is acceptable given gRPC stream safety.
- **Use-after-free in `asyncCallback`** — Adding `return` after `delete` prevents both the dangling pointer capture and the null deref.
- **`const_cast` safety comments** — Good documentation. The cast is safe because the receipt is always a local temporary.

CI is green (cpp-build on ubuntu + macos, build-result, check-license). This is a high-quality, well-tested PR. LGTM.

---

PR #1263, opened 2026-06-09
fix(ci): join git diff output to single line for GITHUB_OUTPUT

lizhimins:
## Summary
- Join `git diff --name-only` output with `tr '\n' ' '` to produce a single-line value for `$GITHUB_OUTPUT`
- Fixes `Invalid format` error in `paths-filter` job on push events, where multi-line output broke the `key=value` format

## Test plan
- [x] Verify `paths-filter` job passes on push to master

🤖 Generated with [Claude Code](https://claude.com/claude-code)


RockteMQ-AI (approved):
### Summary
Fixes the Invalid format error in the paths-filter job by joining multi-line git diff output into a single line with tr. This is the correct fix — GITHUB_OUTPUT requires single-line values. LGTM.

---

PR #1264, opened 2026-06-11
fix(golang): add missing return in SendAsync when producer is off

guyinyou:
When the producer is not running, SendAsync invoked the callback with an error but did not return, causing the subsequent goroutine to also execute. This resulted in the callback being called twice and a goroutine leak.


RongtongJin (approved):

RongtongJin (approved):

RockteMQ-AI (approved):
### Summary
This PR fixes a missing `return` after the error callback in `SendAsync` when the producer is not running. Without it, the callback is invoked twice (once with the error, once from the goroutine) and a goroutine leaks. The fix is correct and minimal.

CI is green for the Go build. LGTM.

---

PR #1265, opened 2026-06-11
[Go] Fix missing return after retry scheduling in process_queue

guyinyou:
In ackMessage0, changeInvisibleDuration, and forwardToDeadLetterQueue0, when the response code is not OK, the *Later() retry function is called but execution falls through to callback(nil), causing the message to be prematurely evicted from cache before the operation actually succeeds. This can lead to message loss. 

Add return statements after each *Later() call to prevent fallthrough, matching the Java implementation's behavior where the future is only resolved on terminal states (success or abandoned).

### DIFF
--- a/golang/process_queue.go
+++ b/golang/process_queue.go
@@ -154,6 +154,7 @@ func (dpq *defaultProcessQueue) forwardToDeadLetterQueue0(mv *MessageView, attem
 			" clientId=%s, consumerGroup=%s, messageId=%s, attempt=%d, mq=%s, endpoints=%v, requestId=%s, status message=[%s]", clientId, consumerGroup, messageId, attempt, dpq.mqstr,
 			endpoints, requestId, status.GetMessage())
 		dpq.forwardToDeadLetterQueueLater(mv, 1+attempt, callback)
+		return
 	}
 	// Set result if succeed in changing invisible time.
 	callback(nil)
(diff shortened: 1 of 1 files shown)


golang/process_queue.go
```diff
@@ -154,6 +154,7 @@ func (dpq *defaultProcessQueue) forwardToDeadLetterQueue0(mv *MessageView, attem
 			" clientId=%s, consumerGroup=%s, messageId=%s, attempt=%d, mq=%s, endpoints=%v, requestId=%s, status message=[%s]", clientId, consumerGroup, messageId, attempt, dpq.mqstr,
 			endpoints, requestId, status.GetMessage())
 		dpq.forwardToDeadLetterQueueLater(mv, 1+attempt, callback)
```
RongtongJin, line 156: This retry path still does not retry forwarding to the DLQ. With the new `return`, the message is no longer evicted immediately after scheduling, but `forwardToDeadLetterQueueLater` currently calls `ackMessage0` when the timer fires, and its recover path schedules `ackMessageLater`. If that ACK succeeds, the callback evicts the message even though it never reached the DLQ. Please make `forwardToDeadLetterQueueLater` re-enter `forwardToDeadLetterQueue0`, and have the recover path reschedule `forwardToDeadLetterQueueLater`.

RongtongJin (commented):
One inline review comment from my manual review.

RongtongJin (approved):

PR #1266, opened 2026-06-11
[Go] Fix forwardToDeadLetterQueueLater incorrectly calling ackMessage instead of forwardToDeadLetterQueue

guyinyou:
forwardToDeadLetterQueueLater was calling ackMessage0 and ackMessageLater instead of forwardToDeadLetterQueue0 and forwardToDeadLetterQueueLater, which would cause messages destined for the dead letter queue to be silently acknowledged and lost.

### DIFF
--- a/golang/process_queue.go
+++ b/golang/process_queue.go
@@ -177,10 +177,10 @@ func (dpq *defaultProcessQueue) forwardToDeadLetterQueueLater(mv *MessageView, a
 			if err := recover(); err != nil {
 				dpq.consumer.cli.log.Errorf("[Bug] Failed to schedule message change invisible duration request, mq=%s, messageId=%s, "+
 					"clientId=%s", dpq.mqstr, messageId, clientId)
-				dpq.ackMessageLater(mv, 1+attempt, callback)
+				dpq.forwardToDeadLetterQueueLater(mv, 1+attempt, callback)
 			}
 		}()
-		dpq.ackMessage0(mv, attempt, callback)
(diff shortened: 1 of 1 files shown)


golang/process_queue.go
```diff
@@ -177,10 +177,10 @@ func (dpq *defaultProcessQueue) forwardToDeadLetterQueueLater(mv *MessageView, a
 			if err := recover(); err != nil {
 				dpq.consumer.cli.log.Errorf("[Bug] Failed to schedule message change invisible duration request, mq=%s, messageId=%s, "+
 					"clientId=%s", dpq.mqstr, messageId, clientId)
-				dpq.ackMessageLater(mv, 1+attempt, callback)
+				dpq.forwardToDeadLetterQueueLater(mv, 1+attempt, callback)
 			}
 		}()
-		dpq.ackMessage0(mv, attempt, callback)
+		dpq.forwardToDeadLetterQueue0(mv, attempt, callback)
```
RongtongJin, line 183: This fixes the retry target, but this PR still leaves the `code != OK` branch in `forwardToDeadLetterQueue0` without a `return`. After scheduling this retry, execution falls through to `callback(nil)`, so callers can evict the message before the DLQ forward actually succeeds. Please include the missing `return` from the retry branch as well, or make this PR depend on/rebase on the PR that adds it.

RongtongJin (commented):
One inline review comment from my manual review.

RongtongJin (approved):

PR #1267, opened 2026-06-11
[C#] Fix missing return after InvalidReceiptHandle in AckMessage and ChangeInvisibleDuration

guyinyou:
Both methods in ProcessQueue.cs had a fall-through bug where the InvalidReceiptHandle branch called tcs.SetException() but did not return, causing execution to continue into the subsequent 'if (statusCode != Code.Ok)' block. This resulted in an unnecessary call to AckMessageLater / ChangeInvisibleDurationLater with an already-faulted TaskCompletionSource, which would throw an InvalidOperationException at runtime.

Add the missing 'return;' statements to match the behavior of the Java reference implementation, where each non-success branch explicitly returns after resolving the future.

### DIFF
--- a/csharp/rocketmq-client-csharp/ProcessQueue.cs
+++ b/csharp/rocketmq-client-csharp/ProcessQueue.cs
@@ -400,6 +400,7 @@ namespace Org.Apache.Rocketmq
                                         $" attempt={attempt}, mq={_mq}, endpoints={endpoints}, requestId={requestId}," +
                                         $" status message={status.Message}");
                         tcs.SetException(new BadRequestException((int)statusCode, requestId, status.Message));
+                        return;
                     }
 
                     if (statusCode != Code.Ok)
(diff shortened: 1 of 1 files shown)


RongtongJin (approved):

oss-sentinel-ai (approved):
## Review: Approved ✅

**PR:** #1267 — [C#] Fix missing return after InvalidReceiptHandle in AckMessage and ChangeInvisibleDuration
**Type:** Bug fix (1 file, +2/-0)

### Assessment
Fixes fall-through bugs in ProcessQueue.cs:
- `AckMessage`: InvalidReceiptHandle branch calls `tcs.SetException()` but continues into `statusCode != Code.Ok` block → `AckMessageLater` called with already-faulted TCS → InvalidOperationException
- `ChangeInvisibleDuration`: Same pattern → `ChangeInvisibleDurationLater` called with faulted TCS

Adding `return;` after `SetException()` matches the Java reference implementation.

### Verdict
✅ Critical fix preventing InvalidOperationException at runtime. Companion to the Go client fixes (#1264, #1265).

---

PR #1268, opened 2026-06-11
[C#] Fix SetBody validation logic to correctly reject null and empty body

guyinyou:
The original condition `null != body || body.Length == 0` had two bugs:
- When body is null, it short-circuits to evaluating `body.Length`, causing NullReferenceException instead of a proper ArgumentException.
- When body is a non-null empty array (byte[0]), `null != body` is true, so the empty body passes validation silently.

Fix to `null != body && body.Length > 0`, consistent with the Java implementation which uses `ArrayUtils.isNotEmpty(body)`.

### DIFF
--- a/csharp/rocketmq-client-csharp/Message.cs
+++ b/csharp/rocketmq-client-csharp/Message.cs
@@ -112,7 +112,7 @@ namespace Org.Apache.Rocketmq
 
             public Builder SetBody(byte[] body)
             {
-                Preconditions.CheckArgument(null != body || body.Length == 0, "body should not be empty");
+                Preconditions.CheckArgument(null != body && body.Length > 0, "body should not be empty");
                 _body = body;
                 return this;
             }


oss-sentinel-ai (approved):
## Review: Approved ✅

**PR:** #1268 — [C#] Fix SetBody validation logic to correctly reject null and empty body
**Type:** Bug fix (1 file, +1/-1)

### Assessment
Fixes a critical logic error in `SetBody` validation:
- **Before:** `null != body || body.Length == 0` — when body is null, short-circuits to `body.Length` → NullReferenceException; when body is empty array, passes validation silently
- **After:** `null != body && body.Length > 0` — correctly validates non-null and non-empty

Consistent with Java's `ArrayUtils.isNotEmpty(body)`.

### Verdict
✅ One-character fix (`||` → `&&`) that prevents both NPE and silent empty-body acceptance.

---

RongtongJin (approved):

PR #1285, opened 2026-06-17
[Go] Support offset option for lite topic subscriptions

### LINKED ISSUE #1284: [Feature] [Go] Support offset option for lite topic subscriptions

Go

The Go LitePushConsumer currently supports subscribing to lite topics, but it does not provide a way to specify the starting offset when calling `SubscribeLite`.

Java clients already support this capability through `OffsetOption`, allowing users to subscribe from a specific offset, tail N messages, timestamp, or predefined policies such as last/min/max offset. Without the same support in Go, users cannot control where lite topic consumption starts, which makes it difficult to replay messages, skip existing messages, or resume from a known lite topic offset.

Add offset option support for Go lite topic subscriptions, aligned with the Java client API and protobuf protocol.

The Go client should provide an `OffsetOption` type with support for:

- `LastOffset`
- `MinOffset`
- `MaxOffset`
- `NewOffsetOptionWithOffset(offset int64)`
- `NewOffsetOptionWithTailN(tailN int64)`
- `NewOffsetOptionWithTimestamp(timestamp int64)`

`LitePushConsumer.SubscribeLite` should accept an optional offset option and send it through `SyncLiteSubscriptionRequest.offset_option`.

The offset behavior should follow the existing protocol semantics used by Java:

- Explicit offset is zero-based and inclusive.
- For example, setting offset to `3` starts consuming from the message whose lite offset is `3`, i.e. the fourth message if offsets are `0, 1, 2, 3`.
- `LastOffset` maps to `Policy.LAST`, consistent with Java.
- `MaxOffset` maps to `Policy.MAX` and can be used to skip existing messages.

Users could manually track offsets outside the client, but without client-side support for sending `offset_option` in lite subscription requests, the server cannot start consumption from the desired lite topic offset.

### Additional Context

_No response_

Alienero:
Add OffsetOption helpers and pass optional offset options through SyncLiteSubscriptionRequest when subscribing to lite topics.

Fixes #1284

### DIFF
--- a/golang/lite_push_consumer.go
+++ b/golang/lite_push_consumer.go
@@ -30,7 +30,7 @@ import (
 
 type LitePushConsumer interface {
 	PushConsumer
-	SubscribeLite(liteTopic string) error
+	SubscribeLite(liteTopic string, offsetOption ...OffsetOption) error
 	UnSubscribeLite(liteTopic string) error
 }
 
@@ -107,11 +107,18 @@ func (lpc *defaultLitePushConsumer) notifyUnsubscribeLite(command *v2.NotifyUnsu
 	lpc.litePushConsumerSettings.liteTopicSet.Delete(liteTopic)
 }
 
-func (lpc *defaultLitePushConsumer) SubscribeLite(liteTopic string) error {
+func (lpc *defaultLitePushConsumer) SubscribeLite(liteTopic string, offsetOption ...OffsetOption) error {
 	if err := lpc.checkRunning(); err != nil {
 		return err
 	}
-	if err := lpc.syncLiteSubscription(context.TODO(), v2.LiteSubscriptionAction_PARTIAL_ADD, []string{liteTopic}); err != nil {
+	if len(offsetOption) > 1 {
+		return errors.New("only one offset option is supported")
+	}
+	var option *OffsetOption
+	if len(offsetOption) == 1 {
+		option = &offsetOption[0]
+	}
+	if err := lpc.syncLiteSubscription(context.TODO(), v2.LiteSubscriptionAction_PARTIAL_ADD, []string{liteTopic}, option); err != nil {
 		sugarBaseLogger.Errorf("LitePushConsumer SubscribeLite liteTopic:%s err:%v", liteTopic, err)
 		return err
 	}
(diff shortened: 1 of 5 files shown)


RockteMQ-AI (commented):

### Summary
This PR adds `OffsetOption` support to `SubscribeLite()` in the Go client SDK, allowing consumers to specify where to start consuming when subscribing to a lite topic. Four offset types are supported: Policy (LAST/MIN/MAX), specific Offset, TailN (last N messages), and Timestamp.

### Findings

- **[Info]** `golang/offset_option.go:28` — Using `int64` for `value` to store both policy enum values (1/2/3) and numeric offsets is pragmatic but sacrifices type safety. A zero-value `OffsetOption{}` would have `optionType=0` and `value=0`, which is neither a valid policy nor a meaningful offset. Consider adding a guard in `toProtobuf()` that returns an error for `optionType == 0` instead of silently returning `nil`.

- **[Info]** `golang/lite_push_consumer.go:113` — The variadic `...OffsetOption` pattern with `len > 1` check is a clean backward-compatible API design. Existing callers of `SubscribeLite("topic")` continue to work without changes. 👍

- **[Warning]** `golang/lite_push_consumer.go:116-119` — Taking `&offsetOption[0]` works correctly, but note that this creates a pointer to the variadic slice element. This is safe in Go (the slice is heap-allocated when its address is taken), but an alternative pattern like accepting a single `*OffsetOption` or using a builder/options struct would be more explicit about the "at most one" constraint at the type level.

- **[Info]** `golang/offset_option.go:75-102` — The `toProtobuf()` method correctly maps each option type to the corresponding protobuf oneof variant. Input validation (negative checks) is consistent across `Offset()`, `TailN()`, and `Timestamp()`.

- **[Info]** `golang/offset_option_test.go` — Test coverage is thorough: all four option types are tested, negative value rejection is verified, and the `SubscribeLite` integration test confirms the option propagates correctly through the mock.

### Suggestions

1. Consider adding a `Validate() error` method on `OffsetOption` or an explicit check in `toProtobuf()` for the zero-value case (`optionType == 0`) to fail fast rather than silently returning nil.
2. The exported sentinel variables (`LastOffset`, `MinOffset`, `MaxOffset`) are well-designed — they're value types that can be passed directly without allocation.

### Cross-repo Note
This change is Go-client-only. If the same offset option feature is planned for Java/C++/Rust clients, the protobuf schema changes in `proto/` should be coordinated to ensure API consistency across languages.

---

Alienero:
hi @RongtongJin , would you mind reviewing this PR when you get a chance? Thanks!

f1amingo (approved):
LGTM

guyinyou (approved):
lgtm

RongtongJin (approved):

PR #1291, opened 2026-07-01
[Go] Fix LitePushConsumer.WrapReceiveMessageRequest using wrong AutoRenew value

f1amingo:
### Summary

`defaultLitePushConsumer.WrapReceiveMessageRequest` was a copy-paste override
that incorrectly set `AutoRenew: false` and omitted `FilterExpression`.

This PR removes the redundant override so that `defaultLitePushConsumer`
delegates to the embedded `*defaultPushConsumer` implementation, which
correctly sets `AutoRenew: true` and includes `FilterExpression`.

Additionally, redundant tests for `WrapReceiveMessageRequest` on
`defaultLitePushConsumer` are removed since the method now delegates
to the parent and is already covered by `TestDefaultPushConsumer_WrapReceiveMessageRequest`.

### Changes

- Remove `WrapReceiveMessageRequest` override from `defaultLitePushConsumer`
- Drop unused imports: `uuid`, `durationpb`
- Fix test assertion: `AutoRenew` should be `true` for lite push consumer
- Delete `TestLitePushConsumer_WrapReceiveMessageRequest` (redundant)
- Rename `TestDefaultLitePushConsumer_Wraps` → `TestDefaultLitePushConsumer_WrapHeartbeatRequest`, keep only the LitePushConsumer-specific `ClientType` assertion

Unit tests updated and aligned with the delegation behavior.

### DIFF
--- a/golang/lite_push_consumer.go
+++ b/golang/lite_push_consumer.go
@@ -24,8 +24,6 @@ import (
 
 	"github.com/apache/rocketmq-clients/golang/v5/pkg/ticker"
 	v2 "github.com/apache/rocketmq-clients/golang/v5/protocol/v2"
-	"github.com/google/uuid"
-	"google.golang.org/protobuf/types/known/durationpb"
 )
 
 type LitePushConsumer interface {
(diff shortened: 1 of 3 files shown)


guyinyou (approved):
lgtm

imzs (approved):

RockteMQ-AI (approved):

### Summary
Fixes a copy-paste bug in `defaultLitePushConsumer.WrapReceiveMessageRequest` where `AutoRenew` was incorrectly set to `false` and `FilterExpression` was omitted. The fix removes the redundant override so the method correctly delegates to the embedded `*defaultPushConsumer` implementation.

### Findings

- **[Info]** `golang/lite_push_consumer.go:196-213` — The removed override had two bugs: (1) `AutoRenew: false` instead of `true`, and (2) missing `FilterExpression` field. Delegation to the parent implementation is the correct approach — cleaner than patching the override, and ensures future changes to the parent method are automatically inherited.

- **[Info]** `golang/lite_push_consumer_test.go` — Removing `TestLitePushConsumer_WrapReceiveMessageRequest` is correct since the method now delegates to parent and is already covered by `TestDefaultPushConsumer_WrapReceiveMessageRequest`. Notably, the old test itself had a bug: the error message said `"expected auto renew to be true"` but the assertion condition triggered when `AutoRenew` *was* true — a clear copy-paste inconsistency that this PR also fixes by removal.

- **[Info]** `golang/push_consumer_test.go` — Renaming `TestDefaultLitePushConsumer_Wraps` → `TestDefaultLitePushConsumer_WrapHeartbeatRequest` improves test clarity by matching the actual test scope. Removing the stale `WrapReceiveMessageRequest` assertions from this test is correct.

- **[Info]** Unused imports (`uuid`, `durationpb`) are properly cleaned up.

### Verdict
Clean, well-scoped fix. The delegation approach is preferable to patching the override. Tests are properly aligned with the new behavior. LGTM 👍

---

RongtongJin (approved):

PR #1297, opened 2026-07-09
[C++] Add CMake install/uninstall support and enable tests in CI

lizhimins:
## Summary

Add CMake install/uninstall support for the C++ client library and enable unit tests in CI workflows.

## Changes

- **CMake install/uninstall**: Add `install()` rules for headers, libraries, and CMake package config files so downstream projects can use `find_package(rocketmq)`. Include `cmake_uninstall.cmake.in` for clean uninstallation.
- **CI**: Enable C++ unit tests in GitHub Actions workflows (`cpp_build.yml`, `cpp_coverage.yml`).
- **Bazel**: Add `.bazelignore` to exclude build output directories.

## Testing

- CMake `cmake --install` produces correct header/library layout
- All existing unit tests pass with both CMake and Bazel builds
- CI workflows execute `ctest` after build


RockteMQ-AI:
- **[Good] SOVERSION / VERSION on shared library** — `VERSION 5.1.0` + `SOVERSION 5` enables proper soname versioning (`librocketmq.so.5 → librocketmq.so.5.1.0`), which is important for downstream package management.

- **[Good] Uninstall target** — Standard `cmake_uninstall.cmake.in` pattern, correctly reverses install order via `list(REVERSE)`.

- **[Suggestion] `rocketmq-config.cmake.in` missing transitive deps** — The config template only has `find_dependency(absl)` and `find_dependency(gRPC)`, but the public interface also links `absl::base` and `gRPC::grpc++`. If protobuf or OpenSSL headers are exposed in the public API (e.g., via `Tracing.h` or gRPC types in headers), consider adding:
  ```cmake
  find_dependency(protobuf CONFIG)
  find_dependency(OpenSSL)
  ```
  Without these, downstream `find_package(rocketmq)` may fail if the consumer doesn't already have them loaded.  - **[Info] `BUILD_TESTS` default changed from `ON` → `OFF`** — This means developers building locally without `-DBUILD_TESTS=ON` will skip tests. CI explicitly enables it, so coverage is maintained. Just noting this as a developer-experience change — existing build scripts that relied on tests being built by default will need updating.  - **[Info] Static library excluded from install** — The comment explains the Windows `.lib` collision rationale. Acceptable, but worth noting that consumers who need static linking will have to build from source.  - **[Info] Hardcoded `VERSION 5.1.0`** — Consider reading the version from a file or the parent project to keep it in sync with releases. Not a blocker for this PR.  ### Verdict  Clean, well-structured CMake modernization. The install/uninstall support and CI test enablement are valuable additions. The `$<BUILD_INTERFACE>` scoping for opencensus is a nice touch. The only actionable suggestion is reviewing whether `rocketmq-config.cmake.in` needs additional `find_dependency()` entries for transitive public deps. REVIEW_EOF echo "Review written to /tmp/review_1297.md"

RockteMQ-AI (commented):

### Summary
Adds CMake install/uninstall support for the C++ client library, enabling downstream projects to use `find_package(rocketmq)`. Also enables unit tests in CI workflows and adds minimum version requirements for dependencies.

### Findings

- **[Info]** `cpp/CMakeLists.txt` — Good practice to pin minimum versions for `protobuf` (3.19), `gRPC` (1.46), and `OpenSSL` (1.1) aligned with the gRPC 1.46.x dependency chain. This prevents silent incompatibilities.

- **[Info]** `cpp/CMakeLists.txt:33` — `BUILD_TESTS` default changed from `ON` to `OFF`. This is reasonable for downstream consumers using `add_subdirectory()` or `find_package()`, but note that developers cloning the repo directly must now explicitly pass `-DBUILD_TESTS=ON`. The CI workflows are updated accordingly, so this is consistent.

- **[Info]** `cpp/CMakeLists.txt` — Static library (`rocketmq`) is intentionally not installed to avoid Windows `.lib` collision with the shared library's import library. This is documented in the code comment and is a correct tradeoff.

- **[Info]** `cpp/cmake/rocketmq-config.cmake.in` — Clean package config template. Uses ``@PACKAGE_INIT@`` and `check_required_components` correctly.

### Suggestions

1. Consider adding a `CPack` configuration or at least a note in documentation about how downstream projects should consume the installed package (e.g., `find_package(rocketmq 5.1.0 REQUIRED)`).
2. The `.bazelignore` file only contains `build` — consider whether other build artifact directories (e.g., `cmake-build-*`) should also be ignored.

### Verdict
Clean infrastructure PR. No correctness concerns. LGTM.

---

PR #1305, opened 2026-07-23
[Go] Fix resource leaks on client close

guyinyou:
## Problem

When Go clients are repeatedly created and destroyed, resources leak:

1. **Goroutine leak**: `GracefulStop()` only called `UnRegisterClient()` but never `shutdown()` on the per-client `ClientManager`, leaving 4 background goroutines (heartbeat, idle check, stats, sync settings) and RPC connections alive forever. This was a regression from #235 which made `ClientManager` per-client but forgot to add the shutdown call.

2. **Telemetry session leak**: `endpointsTelemetryClientTable` sessions were never released on client close, leaking gRPC streaming connections.

3. **Metrics cardinality explosion**: All clients shared global OpenCensus Views with `client_id` as a tag key. Each new client added new tag combinations that were never cleaned up, causing unbounded growth in metrics data.

## Fix

1. **ClientManager shutdown**: Call `shutdown()` instead of `UnRegisterClient()` in `GracefulStop()` to stop background goroutines and close RPC connections.

2. **Telemetry session cleanup**: Release all sessions in `GracefulStop()` before shutting down the client manager. Added nil check in `release()` for cases where the observer was never initialized.

3. **Per-client metrics isolation**: Use `view.NewMeter()` (available since OpenCensus v0.22.4) to give each client its own Meter, Measures, Views, and Exporter. On client close, the Meter is stopped and views are unregistered. The `client_id` tag is preserved in exported metric data — the exported metric names and labels are unchanged.

## Test Results

1000 iterations × 50 concurrent workers, each creating/sending/destroying a producer:

| Metric | Master | Fix |
|--------|--------|-----|
| Goroutine leak | +13,814 | +385 |
| Memory Sys | +342 MB | +17 MB |
| Memory Alloc | +200 MB | +3.5 MB |

Goroutines on fix branch stabilize after ~500 iterations (no linear growth), while master grows linearly at ~1,300 per 100 iterations.

### DIFF
--- a/golang/client.go
+++ b/golang/client.go
@@ -181,10 +181,12 @@ func (cs *defaultClientSession) handleTelemetryCommand(response *v2.TelemetryCom
 func (cs *defaultClientSession) release() {
 	cs.observerLock.Lock()
 	defer cs.observerLock.Unlock()
-	if err := cs.observer.CloseSend(); err != nil {
-		cs.cli.log.Errorf("release defaultClientSession err=%v", err)
+	if cs.observer != nil {
+		if err := cs.observer.CloseSend(); err != nil {
+			cs.cli.log.Errorf("release defaultClientSession err=%v", err)
+		}
+		cs.observer = nil
 	}
-	cs.observer = nil
 }
 func (cs *defaultClientSession) publish(ctx context.Context, common *v2.TelemetryCommand) error {
 	var err error
@@ -646,8 +648,14 @@ func (cli *defaultClient) GracefulStop() error {
 		return fmt.Errorf("client has been closed")
 	}
 	cli.notifyClientTermination()
+	cli.endpointsTelemetryClientsLock.Lock()
+	for _, session := range cli.endpointsTelemetryClientTable {
+		session.release()
+	}
+	cli.endpointsTelemetryClientTable = make(map[string]*defaultClientSession)
+	cli.endpointsTelemetryClientsLock.Unlock()
 	if cli.clientManager != nil {
-		cli.clientManager.UnRegisterClient(cli)
+		cli.clientManager.shutdown()
 	}
 	cli.done <- struct{}{}
 	close(cli.done)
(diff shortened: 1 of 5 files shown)


RockteMQ-AI (commented):

### Summary
Fixes three resource leak issues in the Go client when clients are repeatedly created and destroyed: (1) goroutine leak from missing `ClientManager.shutdown()`, (2) telemetry session leak from unreleased gRPC streams, and (3) OpenCensus metrics cardinality explosion from shared global Views. The benchmark results are compelling — goroutine leak reduced from +13,814 to +385, memory from +342 MB to +17 MB.

### Findings

- **[Warning]** `golang/metric.go` — Adding `record(mt meterType, ...)` to the `ClientMeterProvider` interface is a breaking change for any external implementations. Since this is an internal Go package, this is likely acceptable, but it should be noted in the PR description or commit message.

- **[Warning]** `golang/metric.go` — Measure names are now prefixed with `rocketmq_{clientID}_` (e.g., `rocketmq_client-123_publish_latency`). While the view names (e.g., `rocketmq_send_cost_time`) remain stable and are what gets exported, the underlying per-client measure names are unique. If any downstream tooling or dashboards reference raw measure names rather than view names, they would break. Consider documenting this semantic change.

- **[Info]** `golang/client.go` — The `GracefulStop()` telemetry cleanup correctly acquires the lock, iterates and releases all sessions, then reassigns the map. The nil check added to `release()` (`if cs.observer != nil`) properly handles the case where the observer was never initialized. Good defensive fix.

- **[Info]** `golang/client_manager.go` — Adding `shutdown()` to the `ClientManager` interface is the correct approach. The mock updates in `client_manager_mock.go` and `lite_push_consumer_test.go` are consistent.

### Suggestions
1. Consider adding a brief comment on the `shutdown()` method in the `ClientManager` interface explaining its purpose (stops background goroutines and closes RPC connections), since the method name alone is generic.
2. The per-client `Meter` isolation is a clean solution. If OpenCensus adds a lighter-weight "unregister views" API in the future, this could be simplified.

Overall: Well-structured fix addressing three related resource leaks with convincing benchmark evidence.

---

RongtongJin (approved):

PR #1310, opened 2026-07-23
[Java] Improve gRPC transport resilience

### LINKED ISSUE #1309: [Bug] Java gRPC client should normalize transport throttling and recover half-open channels promptly

- [x] I found a bug, not just asking a question, which should be created in GitHub Discussions.
- [x] I have searched the GitHub Issues and GitHub Discussions of this repository and believe that this is not a duplicate.
- [x] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

Java

Any

Java client `master` / 5.2.1. The problem also affects older versions with the same RPC and channel handling.

### Run or Compiler Version

Any supported JDK

### Describe the Bug

The Java gRPC client currently has two related transport-resilience gaps.

#### 1. gRPC `RESOURCE_EXHAUSTED` is not recognized as throttling

The client recognizes the RocketMQ response-body status `Code.TOO_MANY_REQUESTS` in `StatusChecker` and converts it to `TooManyRequestsException`.

However, if a server or an intermediary returns the canonical gRPC transport status `RESOURCE_EXHAUSTED`, the RPC future fails with `StatusRuntimeException`. The response transformation and `StatusChecker` are never invoked.

As a result, `ProducerImpl` treats the failure as a non-throttling error and retries immediately instead of applying the configured throttling backoff. Push-consumer receive logic also uses the normal failure delay instead of its flow-control delay.

#### 2. A half-open TCP connection may be reused long after heartbeats start failing

`RpcClientImpl` currently configures:

```java
keepAliveTime(300, TimeUnit.SECONDS)
keepAliveTimeout(30, TimeUnit.SECONDS)
keepAliveWithoutCalls(true)
```

The client sends heartbeats every 10 seconds with the configured RPC deadline (3 seconds by default), but `ClientImpl#doHeartbeat` only logs heartbeat failures. It does not ask the cached `ManagedChannel` to create a new transport.

If packets are silently dropped after a TCP connection is established (no FIN/RST), send RPCs and heartbeat RPCs repeatedly time out while subsequent requests continue using the same cached channel. Recovery may wait for the 5-minute gRPC keepalive plus its timeout, or for the operating-system TCP timeout.

#### Transport throttling

1. Make a gRPC server/proxy return `io.grpc.Status.RESOURCE_EXHAUSTED` for `SendMessage`.
2. Publish a message with the Java client.
3. Observe that the failure remains a `StatusRuntimeException` and Producer retries it as a non-throttling failure.

#### Half-open connection

1. Establish a Java client connection and successfully publish messages.
2. Drop packets for the established connection without sending FIN/RST, for example with a firewall `DROP` rule.
3. Continue publishing messages.
4. Observe heartbeat deadline failures and repeated send timeouts while the same channel remains cached.

- gRPC `RESOURCE_EXHAUSTED` should be normalized to the same client semantic as RocketMQ `Code.TOO_MANY_REQUESTS`, so existing throttling backoff logic is reused.
- Consecutive heartbeat transport failures should cause new RPCs to use a fresh transport within the heartbeat detection window, rather than waiting several minutes for TCP/keepalive timeout.
- Throttling and ordinary application errors must not trigger channel recreation.

### What Did You See Instead?

- `RESOURCE_EXHAUSTED` bypasses `StatusChecker`, so it is not recognized as throttling.
- Heartbeat failures only produce log messages and do not accelerate transport recovery.
- Producer retries may continue on the half-open channel until the underlying transport is eventually declared dead.

#### Normalize transport throttling at the RPC boundary

At a common RPC future/client-manager boundary, convert gRPC `RESOURCE_EXHAUSTED` into `TooManyRequestsException` using RocketMQ response code `42900`, the current RPC request ID, and the gRPC description. Preserve the original `StatusRuntimeException` as the cause. Other gRPC status codes should remain unchanged.

#### Use heartbeat as the primary channel-health signal

Avoid rebuilding the channel for every send failure because send failures include throttling, application errors, server latency, and ambiguous `DEADLINE_EXCEEDED` results. Concurrent producers could also cause a channel-recreation storm.

A possible policy per endpoint is:

- Any heartbeat response proves the transport is alive and resets the consecutive-failure count, even if the response contains a non-OK RocketMQ application status.
- A definite heartbeat transport failure such as connection-closed/`UNAVAILABLE` may trigger recovery immediately.
- Two consecutive heartbeat `DEADLINE_EXCEEDED` failures trigger `ManagedChannel#enterIdle()` so new RPCs create a new transport.
- Add a cooldown to avoid repeated reconnect attempts during a server-wide incident.
- `RESOURCE_EXHAUSTED` must be treated as throttling and must not trigger channel recovery.

`enterIdle()` is preferable to replacing the entire `ManagedChannel`: existing RPCs can finish or reach their deadlines, while new RPCs establish a new transport without replacing cached stubs and channel ownership structures.

### Additional Context

Related issue: #872

PR #870 added default Java gRPC keepalive for #872. That prevents an infinite wait, but the current 5-minute interval still leaves a long outage window, and heartbeat failures do not currently accelerate recovery. This issue describes the remaining Java recovery gap and the separate missing transport-status normalization.

qianye1001:

Fixes #1309

### Brief Description

This change improves Java gRPC client transport recovery:

- Normalize transport-level gRPC `RESOURCE_EXHAUSTED` failures for all operations completed through `RpcFuture` to `TooManyRequestsException`. This lets producer sends use the existing throttling backoff, while preserving the original `StatusRuntimeException` as the cause for all callers.
- Recover a potentially half-open transport only from heartbeat failures or an explicit server `ReconnectEndpointsCommand`. Ordinary request failures do not call `enterIdle()`.
- Trigger heartbeat recovery after two consecutive `DEADLINE_EXCEEDED` failures. An `UNAVAILABLE` heartbeat triggers recovery only when `channel.getState(false) == READY`; otherwise gRPC native reconnection remains in control.
- Keep the per-endpoint 30-second cooldown for heartbeat-initiated recovery. An explicit server reconnect command bypasses the time-based cooldown, while all recovery sources share a separate per-endpoint in-progress guard so concurrent calls cannot execute `enterIdle()` together.
- Rebuild the long-lived telemetry stream together with `ManagedChannel.enterIdle()`, then synchronize settings on the renewed stream.
- Clear telemetry reconnect state when the session stops or its endpoints are deprecated, distinguish expected cancellation from other stream failures in logs, and serialize outbound observer writes and terminal calls.
- Propagate TCP proxy accept-loop failures back to the integration test instead of leaving them only on the executor thread.

Successful heartbeats reset the timeout counter, and non-connectivity failures such as `RESOURCE_EXHAUSTED` do not trigger transport recovery. A heartbeat attempt suppressed by cooldown keeps its accumulated failure count so recovery can proceed promptly after the cooldown expires.

Using JDK 11:

```text
mvn -pl test -am -DskipITs test
```

Results:

- Client module: 290 tests, 0 failures, 0 errors, 1 skipped.
- Integration module: 3 tests, 0 failures, 0 errors.
- Checkstyle and SpotBugs passed.

The integration test uses the existing mock gRPC server and a TCP proxy that blackholes both directions without sending FIN/RST:

- On macOS/NIO, the old baseline recovered after about 330 seconds (test time: 333.331 seconds), matching the configured 300-second keepalive plus 30-second keepalive timeout.
- With this change, the same producer scenario recovered after the two heartbeat timeout cycles, established a new TCP connection, and sent successfully in about 20 seconds (20.738 seconds in the latest full run).

Additional Linux `nftables` fault injection with the real producer showed that the timeout is transport-dependent:

- The old producer recovered in about 30.276 seconds with the default epoll transport because gRPC Netty maps the keepalive timeout to `TCP_USER_TIMEOUT`.
- With effective `TCP_USER_TIMEOUT=0`, the old producer recovered in about 329.993 seconds.
- With this change, the first blackholed connection recovered in about 15.034 seconds. A second immediate heartbeat-initiated fault recovered at the intended 30-second cooldown boundary.

The gRPC Netty behavior is implemented in [`NettyClientTransport`](https://github.com/grpc/grpc-java/blob/v1.53.0/netty/src/main/java/io/grpc/netty/NettyClientTransport.java#L239-L248).

### DIFF
--- a/java/client/src/main/java/org/apache/rocketmq/client/java/impl/Client.java
+++ b/java/client/src/main/java/org/apache/rocketmq/client/java/impl/Client.java
@@ -56,6 +56,13 @@ public interface Client {
     boolean isSslEnabled();
 
     /**
+     * Reconnect telemetry to the specified endpoints.
+     *
+     * @param endpoints endpoints to reconnect.
+     */
+    void reconnectTelemetry(Endpoints endpoints);
+
+    /**
      * Send Heartbeat
      *
      * <p> Send a heartbeat to the remote endpoint.
--- a/java/client/src/main/java/org/apache/rocketmq/client/java/impl/ClientImpl.java
+++ b/java/client/src/main/java/org/apache/rocketmq/client/java/impl/ClientImpl.java
@@ -130,7 +130,7 @@ public abstract class ClientImpl extends AbstractIdleService implements Client, 
     private final ReadWriteLock sessionsLock;
 
     private final CompositedMessageInterceptor compositedMessageInterceptor;
-    private boolean receiveReconnect = false;
+    private volatile boolean receiveReconnect = false;
 
     public ClientImpl(ClientConfiguration clientConfiguration, Set<String> topics) {
         this.clientConfiguration = checkNotNull(clientConfiguration, "clientConfiguration should not be null");
@@ -295,6 +295,7 @@ public abstract class ClientImpl extends AbstractIdleService implements Client, 
     @Override
     public void onReconnectEndpointsCommand(Endpoints endpoints, ReconnectEndpointsCommand command) {
         receiveReconnect = true;
+        getClientManager().reconnect(endpoints);
     }
 
     /**
@@ -553,6 +554,23 @@ public abstract class ClientImpl extends AbstractIdleService implements Client, 
 
     public void setReceiveReconnect(boolean receiveReconnect) {
         this.receiveReconnect = receiveReconnect;
+    }
+
+    @Override
+    public void reconnectTelemetry(Endpoints endpoints) {
+        final ClientSessionImpl clientSession;
+        sessionsLock.readLock().lock();
+        try {
+            clientSession = sessionsTable.get(endpoints);
+        } finally {
+            sessionsLock.readLock().unlock();
+        }
+        if (null == clientSession) {
+            log.warn("Failed to rebuild telemetry because client session does not exist, endpoints={}, clientId={}",
+                endpoints, clientId);
+            return;
+        }
+        clientSession.reconnect();
     }
 
     /**
--- a/java/client/src/main/java/org/apache/rocketmq/client/java/impl/ClientImpl.java
+++ b/java/client/src/main/java/org/apache/rocketmq/client/java/impl/ClientImpl.java
@@ -130,7 +130,6 @@ public abstract class ClientImpl extends AbstractIdleService implements Client, 
     private final ReadWriteLock sessionsLock;
 
     private final CompositedMessageInterceptor compositedMessageInterceptor;
-    private volatile boolean receiveReconnect = false;
 
     public ClientImpl(ClientConfiguration clientConfiguration, Set<String> topics) {
         this.clientConfiguration = checkNotNull(clientConfiguration, "clientConfiguration should not be null");
@@ -294,7 +293,6 @@ public abstract class ClientImpl extends AbstractIdleService implements Client, 
 
     @Override
     public void onReconnectEndpointsCommand(Endpoints endpoints, ReconnectEndpointsCommand command) {
-        receiveReconnect = true;
         getClientManager().reconnect(endpoints);
     }
 
@@ -546,14 +544,6 @@ public abstract class ClientImpl extends AbstractIdleService implements Client, 
     @Override
     public ClientId getClientId() {
         return clientId;
-    }
-
-    public boolean isReceiveReconnect() {
-        return receiveReconnect;
-    }
-
-    public void setReceiveReconnect(boolean receiveReconnect) {
-        this.receiveReconnect = receiveReconnect;
     }
 
     @Override
--- a/java/client/src/main/java/org/apache/rocketmq/client/java/impl/ClientManager.java
+++ b/java/client/src/main/java/org/apache/rocketmq/client/java/impl/ClientManager.java
@@ -63,6 +63,13 @@ public abstract class ClientManager extends AbstractIdleService {
     public abstract ScheduledExecutorService getScheduler();
 
     /**
+     * Reconnect to the specified endpoints.
+     *
+     * @param endpoints endpoints to reconnect.
+     */
+    public abstract void reconnect(Endpoints endpoints);
+
+    /**
      * Query topic route asynchronously, the method ensures no throwable.
      *
      * @param endpoints requested endpoints.
--- a/java/client/src/main/java/org/apache/rocketmq/client/java/impl/ClientManagerImpl.java
+++ b/java/client/src/main/java/org/apache/rocketmq/client/java/impl/ClientManagerImpl.java
@@ -279,14 +279,12 @@ public class ClientManagerImpl extends ClientManager {
         final AtomicLong previousRecoveryNanoTime = heartbeatRecoveryNanoTime.putIfAbsent(
             endpoints, recoveryNanoTime);
         if (null != previousRecoveryNanoTime) {
-            while (true) {
-                final long previous = previousRecoveryNanoTime.get();
-                if (now - previous < HEART_BEAT_RECOVERY_COOLDOWN.toNanos()) {
-                    return;
-                }
-                if (previousRecoveryNanoTime.compareAndSet(previous, now)) {
-                    break;
-                }
+            final long previous = previousRecoveryNanoTime.get();
+            if (now - previous < HEART_BEAT_RECOVERY_COOLDOWN.toNanos()) {
+                return;
+            }
+            if (!previousRecoveryNanoTime.compareAndSet(previous, now)) {
+                return;
             }
         }
         log.warn("Try to recover transport after heartbeat failure, endpoints={}, statusCode={}, clientId={}",
--- a/java/client/src/main/java/org/apache/rocketmq/client/java/impl/ClientManagerImpl.java
+++ b/java/client/src/main/java/org/apache/rocketmq/client/java/impl/ClientManagerImpl.java
@@ -65,6 +65,7 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
@@ -114,6 +115,7 @@ public class ClientManagerImpl extends ClientManager {
     private final ReadWriteLock rpcClientTableLock;
     private final ConcurrentMap<Endpoints, Integer> heartbeatFailureAttempts;
     private final ConcurrentMap<Endpoints, AtomicLong> transportRecoveryNanoTime;
+    private final ConcurrentMap<Endpoints, AtomicBoolean> transportRecovering;
 
     /**
      * In charge of all scheduled tasks.
@@ -131,6 +133,7 @@ public class ClientManagerImpl extends ClientManager {
         this.rpcClientTableLock = new ReentrantReadWriteLock();
         this.heartbeatFailureAttempts = new ConcurrentHashMap<>();
         this.transportRecoveryNanoTime = new ConcurrentHashMap<>();
+        this.transportRecovering = new ConcurrentHashMap<>();
         final long clientIndex = client.getClientId().getIndex();
         this.scheduler = new ScheduledThreadPoolExecutor(
             Runtime.getRuntime().availableProcessors(),
@@ -165,6 +168,7 @@ public class ClientManagerImpl extends ClientManager {
                     it.remove();
                     heartbeatFailureAttempts.remove(endpoints);
                     transportRecoveryNanoTime.remove(endpoints);
+                    transportRecovering.remove(endpoints);
                     rpcClient.shutdown();
                     log.info("Rpc client has been idle for a long time, endpoints={}, idleDuration={}, " +
                             "rpcClientMaxIdleDuration={}, clientId={}", endpoints, idleDuration,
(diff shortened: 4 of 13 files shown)


RockteMQ-AI (commented):

### Summary
This PR improves Java gRPC client transport resilience through multiple coordinated changes: normalizing `RESOURCE_EXHAUSTED` to `TooManyRequestsException` for producer throttling, restricting transport recovery to heartbeat failures and explicit server reconnect commands, implementing a consecutive-failure threshold for heartbeat-driven recovery, and rebuilding telemetry streams during `enterIdle()`.

### Findings

- **[Info]** `RpcFuture.java:55-78` — The `normalizeTransportException` method correctly wraps only `RESOURCE_EXHAUSTED` while preserving the original `StatusRuntimeException` as cause via `initCause()`. Consider adding a brief inline comment explaining *why* only `RESOURCE_EXHAUSTED` is normalized (i.e., to let producer sends use the existing throttling backoff path) — this helps future maintainers understand the design intent.

- **[Info]** `ClientManagerImpl.java` — The heartbeat recovery logic triggers after 2 consecutive `DEADLINE_EXCEEDED` failures. Consider extracting the magic number `2` into a named constant (e.g., `HEARTBEAT_FAILURE_THRESHOLD`) for readability. The 30-second per-endpoint cooldown with suppressed failure count accumulation is well-designed — it prevents recovery storms while allowing prompt retry after cooldown expires.

- **[Info]** `ClientSessionImpl.java` — The `reconnect()` method coordinates `enterIdle()` with telemetry stream rebuild. The per-endpoint in-progress guard (atomic CAS) correctly prevents concurrent `enterIdle()` calls. The synchronization of outbound observer writes and terminal calls is a good defensive measure.

- **[Info]** `ClientImpl.java:293-296` — The refactor from flag-based (`receiveReconnect = true`) to action-based (`getClientManager().reconnect(endpoints)`) handling is cleaner and eliminates the polling delay. The removal of `isReceiveReconnect()`/`setReceiveReconnect()` is consistent.

- **[Info]** `RpcClient.java` — Adding `enterIdle()` and `getState()` to the interface is appropriate for giving `ClientManagerImpl` direct control over channel connectivity state.

### Correctness Assessment

The design correctly separates concerns:
- Ordinary request failures → no transport recovery (gRPC native reconnection handles it)
- Heartbeat `DEADLINE_EXCEEDED` (2 consecutive) → transport recovery with cooldown
- Heartbeat `UNAVAILABLE` (when channel is `READY`) → transport recovery
- Explicit server `ReconnectEndpointsCommand` → bypass cooldown, immediate recovery
- `RESOURCE_EXHAUSTED` → normalized to `TooManyRequestsException` (not a connectivity failure)

The `directExecutor()` usage in `Futures.catchingAsync` is safe since the callback only creates an exception object.

### Tests

Comprehensive test coverage:
- `RpcFutureTest` — verifies `RESOURCE_EXHAUSTED` normalization and passthrough of other status codes
- `ClientManagerImplTest` — covers heartbeat threshold, cooldown, concurrent reconnect guard
- `ClientSessionImplTest` — validates telemetry renewal and reconnect sequence
- Integration test with TCP proxy blackhole — demonstrates recovery from ~330s to ~20s

### Suggestions

1. Extract `2` (consecutive heartbeat failure threshold) to a named constant
2. Add brief design-intent comment in `normalizeTransportException`
3. Consider adding a high-level javadoc on `ClientSessionImpl.reconnect()` describing the reconnect sequence

No critical or warning-level issues found. This is a well-structured improvement to transport resilience.

---

RockteMQ-AI (commented):

### Summary
This PR improves Java gRPC client transport recovery by: (1) normalizing `RESOURCE_EXHAUSTED` gRPC failures to `TooManyRequestsException` in `RpcFuture` for consistent backoff behavior, (2) replacing the passive `receiveReconnect` boolean flag with active reconnection logic triggered by heartbeat failures, and (3) adding a `reconnectTelemetry()` method to rebuild client sessions on half-open connections.

### Findings

- **[Critical]** `ClientSessionImpl.java:reconnect()` — When reconnecting, the old gRPC channel/session should be explicitly closed or cancelled before creating a new one. If `grpcClient` or the underlying `ManagedChannel` is not shut down, this could leak resources (threads, sockets) on repeated reconnections. Please verify that the old session resources are properly released.

- **[Warning]** `RpcFuture.java:55-65` — The `RESOURCE_EXHAUSTED` → `TooManyRequestsException` mapping preserves the original `StatusRuntimeException` as the cause, which is good. However, ensure that callers catching `TooManyRequestsException` are aware that `getCause()` may return a `StatusRuntimeException` — this nested exception pattern can be surprising. Consider documenting this in the `RpcFuture` Javadoc.

- **[Warning]** `ClientImpl.java:onReconnectEndpointsCommand()` — The method now calls `getClientManager().reconnect(endpoints)` directly. If this method is called from a gRPC callback thread and `reconnect()` is blocking or long-running, it could block the gRPC event loop. Consider wrapping in an async call or documenting the threading contract.

- **[Info]** `ClientImpl.java:reconnectTelemetry()` — The implementation correctly uses `sessionsLock.readLock()` to look up the session, then calls `clientSession.reconnect()` outside the lock. This avoids holding the lock during the potentially slow reconnection. Good design.

- **[Info]** `ClientManagerImpl.java:reconnect()` — The method iterates over sessions and reconnects matching ones. The use of `sessionsLock.readLock()` is correct since we're only reading the table.

- **[Info]** Integration test `ProducerHalfOpenTcpRecoveryIntegrationTest.java` — The TCP blackhole proxy approach is creative and tests the actual recovery path. The 60s timeout and 40s max recovery time are reasonable. However, the test is quite complex (296 lines) and may be flaky in CI environments with variable network timing. Consider adding `@Disabled` or marking it as an integration test that runs only in specific profiles.

### Suggestions
1. Add a Javadoc note on `RpcFuture` explaining the exception wrapping behavior for `RESOURCE_EXHAUSTED`.
2. Consider adding metrics/counters for reconnection attempts to help diagnose transport instability in production.
3. The integration test could benefit from a `@Tag("integration")` or similar mechanism to exclude it from fast unit test runs.

### Cross-repo Note
This change is client-side only and does not require coordinated changes in `apache/rocketmq`. However, the `RESOURCE_EXHAUSTED` → `TooManyRequestsException` mapping assumes the Proxy/Broker returns `RESOURCE_EXHAUSTED` for throttling. Verify that the server-side status code matches this expectation.

---

RockteMQ-AI (commented):

### Summary
Significant improvement to the Java gRPC client's transport resilience. Adds heartbeat failure monitoring with automatic transport recovery, telemetry stream reconnection on failure, and proper thread-safety around `requestObserver` access in `ClientSessionImpl`.

### Architecture

The change introduces a layered recovery mechanism:
1. **Heartbeat monitoring** (`ClientManagerImpl.monitorHeartbeat`) — tracks consecutive heartbeat failures, triggers recovery after threshold
2. **Transport recovery** (`recoverTransport`) — enters idle mode + reconnects telemetry, with cooldown to prevent thrashing
3. **Telemetry reconnection** (`ClientSessionImpl.reconnect`) — cancels the current stream via `onError(CANCELLED)` and triggers `renewRequestObserver`

### Findings

- **[Critical]** `ClientManagerImpl.java:283-296` — The heartbeat failure classification is well-designed:
  - `UNAVAILABLE` with `READY` state → immediate recovery (transport is broken but channel thinks it's connected)
  - `DEADLINE_EXCEEDED` → count-based threshold (2 consecutive failures)
  - Other errors → reset counter (not transport-related)
  
  This correctly distinguishes transport-level failures from application-level errors.

- **[Warning]** `ClientManagerImpl.java:302-340` — `recoverTransport` uses `AtomicBoolean.compareAndSet` for concurrency control and `TRANSPORT_RECOVERY_COOLDOWN` (30s) to prevent thrashing. The `respectHeartbeatCooldown` flag differentiates server-initiated reconnects (no cooldown) from heartbeat-triggered ones (cooldown). This is correct but the logic is subtle — consider adding a brief Javadoc on `recoverTransport` explaining the two recovery paths.

- **[Warning]** `ClientSessionImpl.java:130-155` — The `reconnect()` method sends `onError(CANCELLED)` to the telemetry stream, which triggers `onError()` → `renewRequestObserver()`. The `reconnecting` AtomicBoolean prevents recursive reconnection. However, the `onError()` callback in `onError(Throwable)` (line 228-243) has a subtle interaction:
  - If `reconnecting.get()` is true and the error is `CANCELLED`, it logs at INFO level (expected)
  - If `reconnecting.get()` is true but the error is NOT `CANCELLED`, it logs at WARN (unexpected failure during reconnect)
  - The `settingsInitFuture.setException(throwable)` path when `!sessionHandler.isRunning()` could race with the reconnect flow. Consider whether the exception should be propagated if we're in the middle of a reconnect.

- **[Info]** `ClientSessionImpl.java:74-120` — Thread-safety improvements: `requestObserver` access is now synchronized via `requestObserverLock`. This fixes potential race conditions between `write()`, `release()`, `renewRequestObserver()`, and `reconnect()`. The `volatile` on `requestObserver` is still useful for reads outside the lock (e.g., null checks in `onError`).

- **[Info]** `ClientManagerImpl.java:217-224` — `getRpcClientIfPresent` returns null instead of throwing, which is correct for the reconnect path where the client might have been removed.

- **[Info]** `RpcClientImpl.java` — `enterIdle()` delegates to `ManagedChannel.enterIdle()` and `getState()` delegates to `ManagedChannel.getState()`. Clean abstraction.

- **[Info]** `ClientImpl.java:133` — Removal of `receiveReconnect` boolean field and replacement with direct `reconnect()` call is cleaner. The old code set a flag that was checked elsewhere; the new code triggers reconnection immediately.

### Suggestions

1. Add Javadoc on `recoverTransport` explaining the `respectHeartbeatCooldown` parameter and the two recovery paths
2. Consider whether `settingsInitFuture.setException()` in `onError()` should be skipped when `reconnecting.get()` is true (the reconnect will create a new session that will set its own `settingsInitFuture`)
3. The `HEART_BEAT_FAILURE_THRESHOLD = 2` and `TRANSPORT_RECOVERY_COOLDOWN = 30s` are hardcoded. Consider making them configurable via `ClientConfiguration` for tuning in different environments

### Verdict

Well-designed transport resilience improvement. The layered recovery mechanism (heartbeat monitoring → transport recovery → telemetry reconnection) is a solid approach. The thread-safety fixes in `ClientSessionImpl` are important. The main concerns are the subtle interaction between reconnect state and error handling, and the hardcoded thresholds.

---

lizhimins (approved):

lizhimins:
Java gRPC 客户端在 TCP 连接变为"半开"（half-open）状态时（如网络黑洞、中间设备静默丢包），恢复时间极长。旧实现依赖 gRPC keepalive 机制（默认 300s keepalive + 30s timeout），导致恢复需要 ~330 秒。

PR #1322, opened 2026-08-07
[ISSUE #1321] [Java] Preserve attempt ID until messages are received

### LINKED ISSUE #1321: [Bug] Java push consumer may change attempt ID before a receive request returns messages

- [x] I found a bug, not just a question.
- [x] I searched existing issues and discussions and did not find a duplicate.
- [x] I confirmed that the behavior is in this client repository.

Java

Linux on Kubernetes. The fault was injected by an external h2c proxy between the Java client and RocketMQ Proxy; the client-to-proxy TCP connection and other HTTP/2 streams remained alive.

- Java client: current `master` at `9abb7f39ce9f74b41873a4f5253ab82af052d7a7`
- Server: RocketMQ 5.x gRPC Proxy/Broker deployment

### Run or Compiler Version

OpenJDK 11

### Describe the Bug

`ProcessQueueImpl` can replace the attempt ID before a receive request has actually returned messages to the client:

1. On a receive RPC failure, the current implementation preserves the request attempt ID only for `DEADLINE_EXCEEDED`. Other transport failures clear it, so the retry uses a new attempt ID.
2. After a successful but empty receive response, the next receive also uses a new attempt ID.

This is observable with FIFO consumption when the server has completed the receive/POP operation but the corresponding HTTP/2 response is lost. The previous attempt may still own the server-side FIFO order state. Retrying with a different attempt ID cannot recover that result and can be blocked by the existing order lock.

An attempt ID should identify consecutive receive retries until a response actually contains messages, rather than identify every client-side RPC invocation.

### Steps to Reproduce

1. Create a new FIFO topic and FIFO consumer group.
2. Start a Java `PushConsumer` and send one FIFO message.
3. Put an external HTTP/2-aware proxy in front of RocketMQ Proxy.
4. After the server sends response DATA for one `ReceiveMessage` RPC, reset only that HTTP/2 stream. Keep the TCP connection and unrelated streams alive.
5. Observe the receive error and the attempt IDs used by subsequent receive requests.
6. Repeat the scenario several times because the timing depends on whether the server has completed POP before the stream is reset.

In a deterministic run, the old client received `CANCELLED: RST_STREAM`, discarded the request attempt ID, generated new IDs for subsequent receives, returned only empty results, and did not invoke the message callback within 180 seconds.

For comparison, a response blackhole that ends as `DEADLINE_EXCEEDED` does not reproduce the same behavior because the old implementation already preserves the attempt ID for that status.

- Reuse the request attempt ID after any failed receive RPC.
- Reuse it after a successful response that contains no messages.
- Generate a new attempt ID only after the raw receive response contains at least one message.
- Preserve the existing acknowledgement and redelivery semantics.

### What Did You See Instead?

After a non-deadline stream failure, the next request used a new attempt ID. In the FIFO scenario above, the new attempt was unable to recover the message associated with the previous attempt and remained blocked behind the existing FIFO order state.

### Additional Context

With the proposed behavior, the retry after the same injected stream reset reused the previous attempt ID, received the original FIFO message in about one second with `deliveryAttempt=1`, and acknowledged it successfully.

The decision to rotate must be based on messages in the raw receive response, before client-side filtering. If an interceptor filters every returned message, the server still returned messages for that attempt and the next receive should use a new ID.

`SimpleConsumer` does not set an attempt ID in its receive request, so its behavior is outside the scope of this bug. This change does not provide exactly-once delivery and does not alter ACK behavior.

qianye1001:

Fixes #1321

### Brief Description

The Java push consumer previously generated a new receive attempt ID after every successful response, including an empty response, and preserved the current attempt ID only when a receive RPC failed with `DEADLINE_EXCEEDED`.

This change keeps the request attempt ID across:

- any failed receive RPC, regardless of the gRPC status; and
- a successful receive response that contains no messages.

A new attempt ID is generated only after the raw receive response contains at least one message. The check intentionally happens before message-interceptor filtering because the server has completed delivery even if the interceptor removes every returned message.

This allows a retry to recover a server-side receive result after a response is lost or one HTTP/2 receive stream is reset, which is especially important when the previous attempt owns FIFO order state.

The change is limited to `PushConsumer`. `SimpleConsumer` does not populate an attempt ID in its receive request, so its behavior is unchanged. ACK, redelivery, and delivery-attempt semantics are not changed, and this does not introduce exactly-once delivery.

Ran the Java test reactor on JDK 11, including compilation, Checkstyle, SpotBugs, and the updated integration test:

```shell
mvn -pl test -am -Dtest=AttemptIdIntegrationTest -Dsurefire.failIfNoSpecifiedTests=false test
```

Result: `BUILD SUCCESS`; `AttemptIdIntegrationTest` passed 1/1.

The integration test verifies one consecutive sequence of a deadline, a non-deadline transport failure, an empty response, and a response containing a message. The first four requests use the same attempt ID, and the request after the message response uses a new ID.

Also validated on Kubernetes with newly created NORMAL/FIFO topics and consumer groups and an external h2c proxy that resets only one `ReceiveMessage` HTTP/2 stream while preserving the TCP connection and unrelated streams:

- Old client: after `CANCELLED: RST_STREAM`, the next receive used a different attempt ID; in the reproduced FIFO lock case no callback occurred within 180 seconds.
- Modified client: the retry reused the same attempt ID, received the original FIFO message in about one second with `deliveryAttempt=1`, and acknowledged it successfully.
- A response blackhole ending in `DEADLINE_EXCEEDED` was used as a control; the old client already retained the attempt ID for that status.

### DIFF
--- a/java/client/src/main/java/org/apache/rocketmq/client/java/impl/consumer/ProcessQueueImpl.java
+++ b/java/client/src/main/java/org/apache/rocketmq/client/java/impl/consumer/ProcessQueueImpl.java
@@ -32,7 +32,6 @@ import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.MoreExecutors;
 import com.google.common.util.concurrent.SettableFuture;
 import com.google.errorprone.annotations.concurrent.GuardedBy;
-import io.grpc.StatusRuntimeException;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -251,6 +250,9 @@ class ProcessQueueImpl implements ProcessQueue {
             Futures.addCallback(future, new FutureCallback<ReceiveMessageResult>() {
                     @Override
                     public void onSuccess(ReceiveMessageResult result) {
+                        // Rotate the attempt ID based on the raw response because messages removed by an interceptor
+                        // were still delivered by the server for this attempt.
+                        final boolean hasReceivedMessages = !result.getMessageViewImpls().isEmpty();
                         // Intercept after message reception.
                         final List<GeneralMessage> generalMessages = result.getMessageViewImpls().stream()
                             .map((Function<MessageView, GeneralMessage>) GeneralMessageImpl::new)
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (commented):
### Summary
Reviewed the changes in this PR. Overall the changes look reasonable.

A few observations from the automated analysis:
- Please ensure adequate test coverage for the changes
- Verify backward compatibility if any public APIs are modified

---

lizhimins (approved):

PR #1324, opened 2026-08-10
[Go] Fix round-robin load balancer always selecting the same queue

guyinyou:
## Summary
- Fix a bug in `TakeMessageQueues` where the index calculation used `next+1` (constant) instead of `next+i` (loop-varying), causing every loop iteration to examine the same message queue
- This defeats the round-robin intent and prevents proper load distribution across brokers
- The same bug exists in both the main loop and the fallback loop

## Test plan
- [ ] Verify round-robin selection distributes across different queues
- [ ] Verify existing tests pass

### DIFF
--- a/golang/loadBalancer.go
+++ b/golang/loadBalancer.go
@@ -73,7 +73,7 @@ func (plb *publishingLoadBalancer) TakeMessageQueues(excluded *sync.Map, count i
 	candidateBrokerNames := make(map[string]bool, 32)
 
 	for i := 0; i < len(plb.messageQueues); i++ {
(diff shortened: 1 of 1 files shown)


RongtongJin (approved):

RockteMQ-AI (approved):
### Summary
Clear bug fix for the Go round-robin load balancer. The original code used `next+1` as the index in both loops, meaning every iteration would select the same message queue regardless of the loop counter. The fix changes it to `next+int32(i)`, correctly advancing through different queues on each iteration.

Minimal, correct, and directly addresses the root cause.

LGTM.

---
*Automated review by github-manager*

PR #1326, opened 2026-08-10
[Java] Support batch acknowledgements for PushConsumer

### LINKED ISSUE #1325: [Feature] Batch acknowledgements for Java gRPC PushConsumer

Java

The Java gRPC PushConsumer currently sends one `AckMessage` RPC for every successfully consumed normal message. The wire protocol already supports repeated `AckMessageRequest.entries`, and RocketMQ Proxy can process multi-entry acknowledgement requests, but the client constructs and sends a one-entry request from `ProcessQueueImpl` for each message.

At high consume rates, the resulting ACK RPC volume consumes a significant amount of client and Proxy CPU without improving delivery semantics.

Add a client-local acknowledgement batcher for Java gRPC PushConsumer with these rules:

- Accumulate acknowledgements independently within each PushConsumer.
- Partition batches by endpoint and topic because one `AckMessageRequest` carries a single topic and is sent to one endpoint.
- Flush a batch when it reaches 1,024 messages or when its oldest entry has waited 5 seconds.
- Only batch messages whose elapsed time from receive/decode to successful consumption is at most 25 seconds. Slow-consumed messages should be acknowledged immediately.
- Keep FIFO acknowledgements immediate because the next ordered message depends on the acknowledgement.
- Keep Lite consumer acknowledgements on the existing immediate path.
- Map per-entry server results back to the corresponding message and preserve the existing individual retry behavior for failed or missing results.
- Flush pending and in-flight acknowledgements during PushConsumer shutdown.

1. **Only enable batch ACK on the server.** This does not reduce RPC volume because the current Java PushConsumer still emits one entry per request.
2. **Batch across topics.** This is incompatible with the protocol because an `AckMessageRequest` has one topic resource.
3. **Delay every acknowledgement, including FIFO messages.** This can block ordered consumption and change existing behavior.
4. **Retry a failed batch as another batch.** Retrying each affected message through the existing single-message path is simpler and preserves current receipt-handle error semantics.

### Additional Context

A remote Kubernetes validation against a commercial RocketMQ deployment with batch ACK enabled used the medium 2B workload: 1,000 topics, 1,000 groups, 4 KiB messages, 2,000 producers, 2,000 consumers, and approximately 8,000 TPS.

In two controlled candidate windows, ACK requests averaged about 20.4 entries per RPC, reducing ACK RPC volume by approximately 95.1%. A high-density single-topic test also observed requests containing exactly 1,024 entries.

In an A-B-B-A comparison at the same throughput, total Java consumer CPU decreased by approximately 20.6%, while total Proxy CPU decreased by approximately 17.7%.

qianye1001:

Fixes #1325

### Brief Description

The Java gRPC PushConsumer currently sends one acknowledgement RPC for every successfully consumed message even though `AckMessageRequest` supports repeated entries. Under a high consume rate, these one-entry requests add substantial client and Proxy overhead.

This change introduces one acknowledgement batcher per PushConsumer:

- Batches are partitioned by endpoint and topic to match the request routing and protocol constraints.
- A batch is flushed after 5 seconds or when it reaches 1,024 messages.
- Only messages consumed within 25 seconds of receive/decode are eligible. Slow messages use the existing immediate acknowledgement path.
- FIFO acknowledgements remain immediate because subsequent ordered messages depend on them.
- Lite consumers retain the existing immediate acknowledgement behavior.
- Multi-entry responses are mapped back to each message by message ID. Missing or failed results use the existing delayed individual retry path; invalid receipt handles preserve the existing no-retry behavior.
- PushConsumer shutdown waits for consumption to finish and then flushes pending and in-flight acknowledgement batches.
- ACK interceptors still receive all messages included in each batch.

The current RocketMQ Proxy protocol and batch-ACK path already accept repeated entries, so this change does not require a protocol update.

At approximately 8,000 TPS in the medium 2B workload, two controlled candidate runs averaged about 20.4 entries per ACK RPC, reducing ACK RPC volume by approximately 95.1%. Total Java consumer CPU decreased by approximately 20.6%, while throughput remained unchanged. A separate single-topic validation observed requests with exactly 1,024 entries.

- `zsh -ic 'javaswitch 11 >/dev/null && mvn -f java/pom.xml -pl client -am test'`
  - 304 tests run
  - 0 failures
  - 0 errors
  - 1 skipped
- Checkstyle: 0 violations
- SpotBugs: 0 errors or warnings
- Verified the compiled batcher class remains Java 8 compatible (`major version: 52`).
- Ran an end-to-end Kubernetes smoke test against a two-Broker/two-Proxy deployment with batch ACK enabled.
- Ran a medium 2B A-B-B-A comparison with 1,000 topics, 1,000 groups, 2,000 producers, 2,000 consumers, 4 KiB messages, 120 seconds warm-up, and 300 seconds sampling per round.

| Metric | Baseline mean | Batch ACK mean | Change |
| --- | ---: | ---: | ---: |
| Send TPS | 7,981.0 | 7,981.5 | +0.01% |
| Consumer total CPU | 3.446 cores | 2.736 cores | -20.59% |
| Consumer CPU p95 | 3.658 cores | 2.975 cores | -18.67% |
| Proxy total CPU | 10.998 cores | 9.052 cores | -17.70% |
| E2E latency p99 | 21.682 ms | 20.515 ms | -5.38% |

### DIFF
--- a/java/client/src/main/java/org/apache/rocketmq/client/java/impl/consumer/AckMessageBatcher.java
+++ b/java/client/src/main/java/org/apache/rocketmq/client/java/impl/consumer/AckMessageBatcher.java
@@ -1,0 +1,292 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.rocketmq.client.java.impl.consumer;
+
+import apache.rocketmq.v2.AckMessageResponse;
+import apache.rocketmq.v2.AckMessageResultEntry;
+import apache.rocketmq.v2.Code;
+import apache.rocketmq.v2.Status;
+import com.google.common.util.concurrent.FutureCallback;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.MoreExecutors;
+import com.google.common.util.concurrent.SettableFuture;
+import com.google.errorprone.annotations.concurrent.GuardedBy;
+import java.time.Duration;
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Deque;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import org.apache.rocketmq.client.java.message.MessageViewImpl;
+import org.apache.rocketmq.client.java.route.Endpoints;
+import org.apache.rocketmq.client.java.rpc.RpcFuture;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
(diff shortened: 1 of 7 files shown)


RockteMQ-AI (approved):
### Summary
Introduces `AckMessageBatcher` to batch PushConsumer acknowledgements by (endpoints, topic), reducing per-message RPC overhead. The design is clean: batches are flushed either when the size threshold is reached or when a delay timer expires. Partial failures (MULTIPLE_RESULTS) are correctly mapped back to individual message futures, and missing results default to INTERNAL_ERROR so the caller can retry. FIFO and lite-consumer messages correctly bypass batching to preserve ordering guarantees.

The retry path (`ackMessageLater`) correctly forces `batchable=false` to prevent re-batching of failed messages, and `flushAndClose()` provides a graceful shutdown mechanism.

Comprehensive test coverage (7 test cases) validates the key scenarios.

LGTM.

---
*Automated review by github-manager*

Issue #1335, opened 2026-08-13
[Bug] [C++] Synchronous producer APIs can block forever; SendMessage has no deadline


lizhimins:
### Before Creating the Bug Report

- [X] I found a bug, not just asking a question, which should be created in [GitHub Discussions](https://github.com/apache/rocketmq-clients/discussions).
- [X] I have searched the [GitHub Issues](https://github.com/apache/rocketmq-clients/issues) of this repository and believe that this is not a duplicate.
- [X] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

### Runtime platform environment

Linux x86_64 / macOS arm64

### RocketMQ description

Apache rocketmq-clients C++ (verified against master `3a8ec615`)

### RocketMQ Client version

cpp-5.0.3, cpp-5.1.0 and current master

### Run or compilation description

Bazel 6.6.0

### Bug description

The synchronous producer APIs can block the calling thread indefinitely. `Configuration::withRequestTimeout()` does not bound them.

**1. `SendMessage` has no gRPC deadline.** `send()` is the only RPC in the `ClientManager` interface without a `timeout` parameter:

| RPC | `timeout` parameter |
| --- | --- |
| `ack(...)` | yes |
| `changeInvisibleDuration(...)` | yes |
| `endTransaction(...)` | yes |
| `recallMessage(...)` | yes |
| `send(...)` | **no** |

Consequently `ClientManagerImpl::send()` (`cpp/source/client/ClientManagerImpl.cpp:288`) never calls `set_deadline()`, and `RpcClientImpl::asyncSend()` passes the context straight to the stub without one. None of the 14 `set_deadline()` call sites in the tree is on the `SendMessage` path. `requestTimeout()` is only consumed by `endTransaction` and `recallMessage` in `ProducerImpl`. A stalled broker therefore leaves the completion callback pending indefinitely.

**2. All four synchronous wait sites use `absl::CondVar::Wait()` with no deadline**, and three also drop wakeups:

| Site | Completion guard | Deadline |
| --- | --- | --- |
| `ProducerImpl::send()` sync (`ProducerImpl.cpp:249`) | correct | **none** |
| `endTransaction0()` (`:463`) | **none at all** | **none** |
| `recallMessage()` (`:579`) | **no flag exists** | **none** |
| `getPublishInfo()` (`:641`) | **racy — read outside the mutex** | **none** |

`getPublishInfo()` is the worst:

```cpp
[...]

Issue #521, opened 2023-05-25
[Bug] [C++] cpp client PushConsumer Crash When dtor


chaojizhuoge:
### Before Creating the Bug Report

- [X] I found a bug, not just asking a question, which should be created in [GitHub Discussions](https://github.com/apache/rocketmq-clients/discussions).

- [X] I have searched the [GitHub Issues](https://github.com/apache/rocketmq-clients/issues) and [GitHub Discussions](https://github.com/apache/rocketmq-clients/discussions)  of this repository and believe that this is not a duplicate.

- [X] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

### Programming Language of the Client

C++

### Runtime Platform Environment

OS: Windows Server 2012 / SUSE Linux Enterprise Server 12 SP5

### RocketMQ Version of the Client/Server

version: rocketmq-clients cpp 5.0.0

### Describe the Bug

[https://github.com/apache/rocketmq-clients/issues/421#issue-1633803209](url)

Same Problem with this closed issue : second coredump

```
(gdb) bt
#0  0x00007fc706eafe5c in __pthread_kill_implementation () from /lib64/libc.so.6
#1  0x00007fc706e5fa76 in raise () from /lib64/libc.so.6
#2  0x00007fc706e497fc in abort () from /lib64/libc.so.6
#3  0x00007fc7070a2b97 in __gnu_cxx::__verbose_terminate_handler() [clone .cold] () from /lib64/libstdc++.so.6
#4  0x00007fc7070ae48c in __cxxabiv1::__terminate(void (*)()) () from /lib64/libstdc++.so.6
#5  0x00007fc7070ae4f7 in std::terminate() () from /lib64/libstdc++.so.6
#6  0x000000000047aafd in std::thread::~thread() ()
#7  0x00000000005b45ff in void std::_Destroy<std::thread>(std::thread*) ()
#8  0x00000000005b3b85 in void std::_Destroy_aux<false>::__destroy<std::thread*>(std::thread*, std::thread*) ()
#9  0x00000000005b2686 in void std::_Destroy<std::thread*>(std::thread*, std::thread*) ()
#10 0x00000000005b048f in void std::_Destroy<std::thread*, std::thread>(std::thread*, std::thread*, std::allocator<std::thread>&) ()
#11 0x00000000005aeffd in std::vector<std::thread, std::allocator<std::thread> >::~vector() ()
[...]

PR #1336, opened 2026-08-13
[ISSUE #1334] [C++] Fix PushConsumer teardown crash and add explicit shutdown()

### LINKED ISSUE #421: cpp client crashed
Modify ExamplePushConsumer.cpp as below, 
```
#include <chrono>
#include <iostream>
#include <mutex>
#include <thread>

#include "gflags/gflags.h"
#include "rocketmq/Logger.h"
#include "rocketmq/PushConsumer.h"

using namespace ROCKETMQ_NAMESPACE;

DEFINE_string(topic, "standard_topic_sample", "Topic to which messages are published");
DEFINE_string(access_point, "121.196.167.124:8081", "Service access URL, provided by your service provider");
DEFINE_string(group, "CID_standard_topic_sample", "GroupId, created through your instance management console");
DEFINE_string(access_key, "", "Your access key ID");
DEFINE_string(access_secret, "", "Your access secret");

void test() {
  std::string tag = "*";

  auto listener = [](const Message& message) {
    std::cout << "Received a message[topic=" << message.topic() << ", MsgId=" << message.id() << "]" << std::endl;
    return ConsumeResult::SUCCESS;
  };

  CredentialsProviderPtr credentials_provider;
  if (!FLAGS_access_key.empty() && !FLAGS_access_secret.empty()) {
    credentials_provider = std::make_shared<StaticCredentialsProvider>(FLAGS_access_key, FLAGS_access_secret);
  }

  auto push_consumer = PushConsumer::newBuilder()
                           .withGroup(FLAGS_group)
                           .withConfiguration(Configuration::newBuilder()
                                                  .withEndpoints(FLAGS_access_point)
                                                  .withRequestTimeout(std::chrono::seconds(3))
                                                  .withCredentialsProvider(credentials_provider)
                                                  .build())
                           .withConsumeThreads(4)
                           .withListener(listener)
                           .subscribe(FLAGS_topic, tag)
                           .build();

  std::this_thread::sleep_for(std::chrono::seconds(3));
}

int main(int argc, char* argv[]) {
  gflags::ParseCommandLineFlags(&argc, &argv, true);

  auto& logger = getLogger();
  logger.setConsoleLevel(Level::Info);
  logger.setLevel(Level::Info);
  logger.init();

  test();
  return EXIT_SUCCESS;
}
```
sometimes, the following core dump was printed,
```
(gdb) bt
#0  0x00000000005984f7 in rocketmq::ReceiveMessageStreamReader::OnReadDone(bool) ()
#1  0x000000000064ecd4 in grpc::internal::ClientCallbackReaderImpl<apache::rocketmq::v2::ReceiveMessageResponse>::StartCall()::{lambda(bool)#2}::operator()(bool) const ()
#2  0x000000000065b514 in void std::__invoke_impl<void, grpc::internal::ClientCallbackReaderImpl<apache::rocketmq::v2::ReceiveMessageResponse>::StartCall()::{lambda(bool)#2}&, bool>(std::__invoke_other, grpc::internal::ClientCallbackReaderImpl<apache::rocketmq::v2::ReceiveMessageResponse>::StartCall()::{lambda(bool)#2}&, bool&&) ()
#3  0x000000000065aa12 in std::enable_if<std::__and_<std::is_void<void>, std::__is_invocable<grpc::internal::ClientCallbackReaderImpl<apache::rocketmq::v2::ReceiveMessageResponse>::StartCall()::{lambda(bool)#2}&, bool> >::value, void>::type std::__invoke_r<void, grpc::internal::ClientCallbackReaderImpl<apache::rocketmq::v2::ReceiveMessageResponse>::StartCall()::{lambda(bool)#2}&, bool>(grpc::internal::ClientCallbackReaderImpl<apache::rocketmq::v2::ReceiveMessageResponse>::StartCall()::{lambda(bool)#2}&, bool&&) ()
#4  0x0000000000659b5d in std::_Function_handler<void (bool), grpc::internal::ClientCallbackReaderImpl<apache::rocketmq::v2::ReceiveMessageResponse>::StartCall()::{lambda(bool)#2}>::_M_invoke(std::_Any_data const&, bool&&) ()
#5  0x0000000000523272 in std::function<void (bool)>::operator()(bool) const ()
#6  0x0000000000522351 in void grpc::internal::CatchingCallback<std::function<void (bool)>&, bool&>(std::function<void (bool)>&, bool&) ()
#7  0x00000000005209ec in grpc::internal::CallbackWithSuccessTag::Run(bool) ()
#8  0x000000000052092b in grpc::internal::CallbackWithSuccessTag::StaticRun(grpc_completion_queue_functor*, int) ()
#9  0x000000000067a9a1 in grpc::(anonymous namespace)::CallbackAlternativeCQ::Ref()::{lambda(void*)#1}::operator()(void*) const ()
#10 0x000000000067a9cf in grpc::(anonymous namespace)::CallbackAlternativeCQ::Ref()::{lambda(void*)#1}::_FUN(void*) ()
#11 0x0000000000ce348c in grpc_core::(anonymous namespace)::ThreadInternalsPosix::ThreadInternalsPosix(char const*, void (*)(void*), void*, bool*, grpc_core::Thread::Options const&)::{lambda(void*)#1}::operator()(void*) const ()
#12 0x0000000000ce34d5 in grpc_core::(anonymous namespace)::ThreadInternalsPosix::ThreadInternalsPosix(char const*, void (*)(void*), void*, bool*, grpc_core::Thread::Options const&)::{lambda(void*)#1}::_FUN(void*) ()
#13 0x00007feef50ae12d in start_thread () from /lib64/libc.so.6
#14 0x00007feef512fbc0 in clone3 () from /lib64/libc.so.6
(gdb)

```
while it may crashed with another core dump,

```
(gdb) bt
#0  0x00007fc706eafe5c in __pthread_kill_implementation () from /lib64/libc.so.6
#1  0x00007fc706e5fa76 in raise () from /lib64/libc.so.6
#2  0x00007fc706e497fc in abort () from /lib64/libc.so.6
#3  0x00007fc7070a2b97 in __gnu_cxx::__verbose_terminate_handler() [clone .cold] () from /lib64/libstdc++.so.6
#4  0x00007fc7070ae48c in __cxxabiv1::__terminate(void (*)()) () from /lib64/libstdc++.so.6
#5  0x00007fc7070ae4f7 in std::terminate() () from /lib64/libstdc++.so.6
#6  0x000000000047aafd in std::thread::~thread() ()
#7  0x00000000005b45ff in void std::_Destroy<std::thread>(std::thread*) ()
#8  0x00000000005b3b85 in void std::_Destroy_aux<false>::__destroy<std::thread*>(std::thread*, std::thread*) ()
#9  0x00000000005b2686 in void std::_Destroy<std::thread*>(std::thread*, std::thread*) ()
#10 0x00000000005b048f in void std::_Destroy<std::thread*, std::thread>(std::thread*, std::thread*, std::allocator<std::thread>&) ()
#11 0x00000000005aeffd in std::vector<std::thread, std::allocator<std::thread> >::~vector() ()
#12 0x00000000005c504c in rocketmq::ThreadPoolImpl::~ThreadPoolImpl() ()
#13 0x00000000005c5094 in rocketmq::ThreadPoolImpl::~ThreadPoolImpl() ()
#14 0x00000000004c885a in std::default_delete<rocketmq::ThreadPool>::operator()(rocketmq::ThreadPool*) const ()
#15 0x00000000004c8230 in std::unique_ptr<rocketmq::ThreadPool, std::default_delete<rocketmq::ThreadPool> >::~unique_ptr() ()
#16 0x00000000004c9d62 in rocketmq::ConsumeMessageServiceImpl::~ConsumeMessageServiceImpl() ()
#17 0x000000000049e1e9 in void std::_Destroy<rocketmq::ConsumeMessageServiceImpl>(rocketmq::ConsumeMessageServiceImpl*) ()
#18 0x000000000049dff0 in void std::allocator_traits<std::allocator<void> >::destroy<rocketmq::ConsumeMessageServiceImpl>(std::allocator<void>&, rocketmq::ConsumeMessageServiceImpl*) ()
#19 0x000000000049d6db in std::_Sp_counted_ptr_inplace<rocketmq::ConsumeMessageServiceImpl, std::allocator<void>, (__gnu_cxx::_Lock_policy)2>::_M_dispose() ()
#20 0x000000000041f54f in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release_last_use() ()
#21 0x000000000041f528 in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release_last_use_cold() ()
#22 0x000000000041ee68 in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release() ()
#23 0x000000000041f6f7 in std::__shared_count<(__gnu_cxx::_Lock_policy)2>::~__shared_count() ()
#24 0x000000000047c8fc in std::__shared_ptr<rocketmq::ConsumeMessageService, (__gnu_cxx::_Lock_policy)2>::~__shared_ptr() ()
#25 0x000000000047c918 in std::shared_ptr<rocketmq::ConsumeMessageService>::~shared_ptr() ()
#26 0x000000000046fb82 in rocketmq::PushConsumerImpl::~PushConsumerImpl() ()
#27 0x000000000046f0bf in void std::_Destroy<rocketmq::PushConsumerImpl>(rocketmq::PushConsumerImpl*) ()
#28 0x000000000046eb5e in void std::allocator_traits<std::allocator<void> >::destroy<rocketmq::PushConsumerImpl>(std::allocator<void>&, rocketmq::PushConsumerImpl*) ()
#29 0x000000000046aa93 in std::_Sp_counted_ptr_inplace<rocketmq::PushConsumerImpl, std::allocator<void>, (__gnu_cxx::_Lock_policy)2>::_M_dispose() ()
#30 0x000000000041f54f in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release_last_use() ()
#31 0x000000000041f528 in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release_last_use_cold() ()
#32 0x000000000041ee68 in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release() ()
#33 0x000000000041f6f7 in std::__shared_count<(__gnu_cxx::_Lock_policy)2>::~__shared_count() ()
#34 0x000000000041f0d8 in std::__shared_ptr<rocketmq::PushConsumerImpl, (__gnu_cxx::_Lock_policy)2>::~__shared_ptr() ()
#35 0x000000000041f0f4 in std::shared_ptr<rocketmq::PushConsumerImpl>::~shared_ptr() ()
#36 0x000000000041f4ae in rocketmq::PushConsumer::~PushConsumer() ()
#37 0x000000000041ded3 in test() ()
#38 0x000000000041e0be in main ()
(gdb)
```

### LINKED ISSUE #1334: [Bug] [C++] PushConsumer crashes during teardown and has no graceful shutdown API

- [X] I have searched the [GitHub Issues](https://github.com/apache/rocketmq-clients/issues) of this repository and believe that this is not a duplicate.

### Runtime platform environment

macOS arm64, Bazel opt build

### RocketMQ description

Apache rocketmq-clients C++ (verified against master `3a8ec615`)

### RocketMQ Client version

cpp-5.1.0 and current master

### Run or compilation description

Bazel 6.6.0, opt build

### Bug description

The public `PushConsumer` API has no `shutdown()` / `close()`; teardown relies entirely on releasing the last `shared_ptr`. When an application releases its last reference while a message is still being consumed, the process crashes with `EXC_BAD_ACCESS` (`KERN_INVALID_ADDRESS` at `0x18`) on the SDK consume worker thread.

Observed stack:

```
spdlog::logger::log(...)
rocketmq::PushConsumerImpl::~PushConsumerImpl()
rocketmq::ConsumeTask::process()
rocketmq::ThreadPoolImpl worker
```

**Root cause.** `ConsumeTask::process()` promotes the service's `weak_ptr` to an *owning* `shared_ptr` on the consume worker thread:

```cpp
// cpp/source/rocketmq/ConsumeTask.cpp:138
std::shared_ptr<PushConsumerImpl> consumer = svc->consumer().lock();
```

`ConsumeMessageServiceImpl` holds the consumer only weakly (`ConsumeMessageService.h:65`), and `ClientManagerImpl` keeps clients as `std::vector<std::weak_ptr<Client>>`, so the application's reference really is the only strong one. Once it is dropped mid-consumption, the local `consumer` above becomes the last owner and `~PushConsumerImpl()` runs on the worker thread when `process()` returns.

Two independent failures follow from that:

1. `~PushConsumerImpl()` logs before doing anything else, so it may touch the static spdlog default logger after it has been destroyed during process teardown — this is the observed `0x18` fault:
   ```cpp
   // cpp/source/rocketmq/PushConsumerImpl.cpp:45
   PushConsumerImpl::~PushConsumerImpl() {
     SPDLOG_DEBUG("DefaultMQPushConsumerImpl is destructed");
     shutdown();
   }
   ```
2. If that log survives, `shutdown()` reaches `ThreadPoolImpl::shutdown()`, which joins **every** thread unconditionally, including the calling worker. Self-join raises `std::system_error(EDEADLK)`; because `PushConsumerImpl::shutdown()` is `noexcept`, that terminates the process. Reproduced in a unit test as `terminate called without an active exception`.

Separately, `ConsumeTask::process()` dereferences `consumer` without a null check, so an already-expired consumer is a third crash path.

The crash is **not FIFO-specific**: `ConsumeTask::process()`, `~PushConsumerImpl()` and `ThreadPoolImpl::shutdown()` are shared by all message types; `fifo_` only selects a `NextStep` branch. FIFO merely widens the window because consumption is serialized.

**Why the official example does not hit this.** `ExamplePushConsumer` sleeps 30 minutes and returns, so destruction happens (a) on the main thread, (b) while no consume task is in flight, and (c) before static destructors run. All three hold only by accident and none is enforced or documented by the API.

For comparison, the other SDKs do expose explicit shutdown — Java's `PushConsumer extends Closeable` with `close()`, Go's `GracefulStop()`. C++ is the only one missing it, which is why this keeps resurfacing.

Prior reports were auto-closed as stale rather than fixed: #421, #521.

### Steps to reproduce

1. Start a `PushConsumer` with a listener returning `ConsumeResult::SUCCESS`.
2. From the application thread, destroy the `PushConsumer` right after the listener signals completion (i.e. while `ConsumeTask::process()` is still on the worker stack).
3. The process faults on the consume worker thread.

### What did you expect to see?

A supported way to shut the consumer down deterministically from the owner thread, draining in-flight consume tasks and ACK callbacks, with the final destruction happening on the owner thread.

### What did you see instead?

`SIGSEGV` (exit 139) on the SDK consume worker thread after the message was successfully consumed and acked.

### Additional context

I have a fix and will open a PR referencing this issue.

lizhimins:

Fixes #1334

Prior reports of the same crash were auto-closed as stale rather than fixed: #421, #521.

### Brief Description

`PushConsumer` has no public `shutdown()` / `close()`, so teardown depends on releasing the last `shared_ptr`. `ConsumeTask::process()` promotes the service's `weak_ptr` into an *owning* `shared_ptr` on a consume worker thread, so an application that drops its last reference while a message is being consumed makes that worker the last owner. `~PushConsumerImpl()` then runs on the worker thread, concurrently with `process()` and process/static teardown, and crashes inside spdlog (`EXC_BAD_ACCESS` at `0x18` on macOS).

This PR:

- **Adds `PushConsumer::shutdown()`.** Called from the owner thread it cancels the periodic tasks and joins the consume workers synchronously. Because the join blocks the owner until in-flight consume tasks finish, a worker can never be the last owner, so the final destruction always happens on the owner thread. It is idempotent and remains safe to call again from the destructor.
- **Stops `ThreadPoolImpl::shutdown()` from joining the calling thread.** Self-join raises `std::system_error(EDEADLK)`, and since `PushConsumerImpl::shutdown()` is `noexcept` that terminated the process. The current worker is now detached instead, so it unwinds once `io_context::run()` returns.
- **Removes the `SPDLOG_DEBUG` call from `~PushConsumerImpl()`**, which could touch the static default logger after it was destroyed.
- **Null-checks the locked consumer in `ConsumeTask::process()`** so an expired `PushConsumerImpl` no longer causes a null dereference.

Two notes for reviewers, so the scope is not overstated:

1. The `ThreadPoolImpl` self-join guard only *hardens* the unsupported path where teardown is driven from a worker thread; it downgrades a guaranteed `terminate` to a survivable unwind. Deterministic teardown still requires calling `PushConsumer::shutdown()` from a non-worker thread, and that is what the API documentation now states.
2. This aligns C++ with the other SDKs rather than adding a new concept — Java already has `PushConsumer extends Closeable` with `close()`, and Go has `GracefulStop()`. C++ was the only one without an explicit shutdown.

The crash is not FIFO-specific: `ConsumeTask::process()`, `~PushConsumerImpl()` and `ThreadPoolImpl::shutdown()` are shared by all message types, and `fifo_` only selects a `NextStep` branch. FIFO just widens the window because consumption is serialized.

Three regression tests were added, and the first two were confirmed to fail against the unfixed code before the fix was applied:

- `ThreadPoolTest.shutdownFromWorkerThreadDoesNotThrowTest` — shuts the pool down from one of its own workers. Before the fix this aborts with `terminate called without an active exception`.
- `ConsumeTaskTest.processWithExpiredConsumerTest` — processes a task whose owning consumer has expired. Before the fix this segfaults.
- `PushConsumerImplTest.shutdownIsIdempotentTest` — covers repeated `shutdown()` calls plus the implicit one from the destructor.

Full C++ suite on this branch, rebased directly on `master`:

```
$ bazel test //source/...
Executed 36 out of 37 tests: 37 tests pass.
```

This branch contains only this fix and sits directly on `master`, so it can be reviewed and merged independently of my producer-side fix for #1335.

### DIFF
--- a/cpp/include/rocketmq/PushConsumer.h
+++ b/cpp/include/rocketmq/PushConsumer.h
@@ -41,6 +41,23 @@ public:
 
   void unsubscribe(const std::string& topic) noexcept;
 
+  /**
+   * Gracefully shut the consumer down.
+   *
+   * Call this from the application (owner) thread before releasing the
+   * PushConsumer. It cancels periodic tasks, then drains and joins the consume
+   * worker threads synchronously on the calling thread. Because the join blocks
+   * the owner until all in-flight consume tasks finish, a consume worker can
+   * never become the last owner of the underlying implementation, so the final
+   * destruction always happens on the owner thread rather than on a worker
+   * thread concurrently with process/static teardown.
+   *
+   * Idempotent: safe to call multiple times and safe to call again from the
+   * destructor. If never called explicitly, teardown still runs from the
+   * destructor, but then it is not guaranteed to run on the owner thread.
+   */
+  void shutdown() noexcept;
+
 private:
   friend class PushConsumerBuilder;
 
--- a/cpp/source/base/ThreadPoolImpl.cpp
+++ b/cpp/source/base/ThreadPoolImpl.cpp
@@ -79,10 +79,24 @@ void ThreadPoolImpl::shutdown() {
   if (state_.compare_exchange_strong(expected, State::STOPPING, std::memory_order_relaxed)) {
     work_guard_->reset();
     context_.stop();
+    auto current_id = std::this_thread::get_id();
     for (auto& thread : threads_) {
-      if (thread.joinable()) {
-        thread.join();
+      if (!thread.joinable()) {
+        continue;
       }
+      if (thread.get_id() == current_id) {
+        // shutdown() was invoked from within one of our own worker threads,
+        // e.g. the last shared_ptr to the owning PushConsumerImpl was released
+        // on a consume worker, so ~PushConsumerImpl -> shutdown() runs here.
+        // Joining the current thread raises std::system_error(EDEADLK); on the
+        // noexcept teardown path that would std::terminate. Detach self so the
+        // worker unwinds naturally once io_context::run() returns. Deterministic
+        // teardown must be driven from a non-worker thread via PushConsumer::shutdown().
+        SPDLOG_WARN("ThreadPool::shutdown() invoked from a worker thread; detaching self to avoid self-join");
+        thread.detach();
+        continue;
+      }
+      thread.join();
     }
     state_.store(State::STOPPED, std::memory_order_relaxed);
   }
--- a/cpp/source/base/tests/ThreadPoolTest.cpp
+++ b/cpp/source/base/tests/ThreadPoolTest.cpp
@@ -20,6 +20,7 @@
 #include "rocketmq/RocketMQ.h"
 #include "gtest/gtest.h"
 #include <atomic>
+#include <chrono>
 #include <functional>
 #include <thread>
 
@@ -71,6 +72,36 @@ TEST_F(ThreadPoolTest, testBasics) {
       cv.Wait(&mtx);
     }
   }
+}
+
+// Regression: shutting the pool down from within one of its own worker threads
+// must not attempt to join the calling thread (self-join deadlocks and raises
+// std::system_error EDEADLK). This mirrors the PushConsumer teardown race where
+// ~PushConsumerImpl runs on a consume worker and drives ThreadPoolImpl::shutdown().
+TEST_F(ThreadPoolTest, shutdownFromWorkerThreadDoesNotThrowTest) {
+  std::atomic<bool> threw{false};
+  std::atomic<bool> finished{false};
+
+  pool_->submit([&]() {
+    try {
+      pool_->shutdown();
+    } catch (...) {
+      threw.store(true);
+    }
+    finished.store(true);
+  });
+
+  auto deadline = std::chrono::steady_clock::now() + std::chrono::seconds(5);
+  while (!finished.load() && std::chrono::steady_clock::now() < deadline) {
+    std::this_thread::sleep_for(std::chrono::milliseconds(10));
+  }
+
+  EXPECT_TRUE(finished.load());
+  EXPECT_FALSE(threw.load());
+
+  // Give the detached worker a moment to unwind out of io_context::run()
+  // before the fixture destroys the pool (test-only synchronization).
+  std::this_thread::sleep_for(std::chrono::milliseconds(200));
 }
 
 ROCKETMQ_NAMESPACE_END
--- a/cpp/source/rocketmq/ConsumeTask.cpp
+++ b/cpp/source/rocketmq/ConsumeTask.cpp
@@ -136,6 +136,12 @@ void ConsumeTask::process() {
   }
 
   std::shared_ptr<PushConsumerImpl> consumer = svc->consumer().lock();
+  if (!consumer) {
+    // The owning PushConsumerImpl has been destructed. Bail out before touching
+    // consumer stats/config so we never dereference a null consumer.
+    SPDLOG_DEBUG("PushConsumer has been destructed; skip processing");
+    return;
+  }
 
   auto self = shared_from_this();
 
--- a/cpp/source/rocketmq/PushConsumer.cpp
+++ b/cpp/source/rocketmq/PushConsumer.cpp
@@ -41,6 +41,16 @@ void PushConsumer::unsubscribe(const std::string& topic) noexcept {
   }
 }
 
+void PushConsumer::shutdown() noexcept {
+  try {
+    if (impl_) {
+      impl_->shutdown();
+    }
+  } catch (const std::exception& e) {
+    SPDLOG_ERROR("Exception in shutdown: {}", e.what());
+  }
+}
+
 PushConsumerBuilder PushConsumer::newBuilder() {
   return {};
 }
--- a/cpp/source/rocketmq/PushConsumerImpl.cpp
+++ b/cpp/source/rocketmq/PushConsumerImpl.cpp
@@ -42,7 +42,10 @@ PushConsumerImpl::PushConsumerImpl(absl::string_view group_name) : ClientImpl(gr
 }
 
 PushConsumerImpl::~PushConsumerImpl() {
-  SPDLOG_DEBUG("DefaultMQPushConsumerImpl is destructed");
+  // Do NOT log here. The destructor may run during process/static teardown when
+  // the static spdlog default logger has already been destroyed; logging then
+  // dereferences a dangling logger (observed as EXC_BAD_ACCESS at 0x18 on macOS).
+  // Deterministic teardown should be driven earlier via PushConsumer::shutdown().
   shutdown();
 }
 
--- a/cpp/source/rocketmq/tests/ConsumeTaskTest.cpp
+++ b/cpp/source/rocketmq/tests/ConsumeTaskTest.cpp
@@ -162,6 +162,33 @@ TEST_F(ConsumeTaskTest, processWithEmptyMessagesTest) {
   task->process();
 }
 
+// Regression: the service is still alive but the owning PushConsumerImpl has
+// already been destructed, so consumer().lock() yields null. process() must not
+// dereference the null consumer (metrics/stats access), it must bail out before
+// invoking the listener.
+TEST_F(ConsumeTaskTest, processWithExpiredConsumerTest) {
+  auto msg = buildMessage("topic", "body");
+  auto task = std::make_shared<ConsumeTask>(weak_service_, weak_pq_, msg);
+
+  bool listener_invoked = false;
+  MessageListener listener = [&listener_invoked](const Message&) {
+    listener_invoked = true;
+    return ConsumeResult::SUCCESS;
+  };
+
+  // Consumer weak_ptr is expired (owning PushConsumerImpl already gone).
+  EXPECT_CALL(*service_, consumer()).WillRepeatedly(testing::Return(std::weak_ptr<PushConsumerImpl>()));
+  ON_CALL(*service_, listener()).WillByDefault(testing::ReturnRef(listener));
+
+  // With no consumer, process() must not run the listener or ack/nack.
+  EXPECT_CALL(*service_, preHandle(testing::_)).Times(0);
+  EXPECT_CALL(*service_, ack(testing::_, testing::_)).Times(0);
+  EXPECT_CALL(*service_, nack(testing::_, testing::_)).Times(0);
+
+  task->process();  // must not dereference a null consumer
+  EXPECT_FALSE(listener_invoked);
+}
+
 // --- process() state-machine: Consume → Ack on SUCCESS ---
 
 TEST_F(ConsumeTaskTest, processConsumeSuccessCallsAckTest) {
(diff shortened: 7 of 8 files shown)


RockteMQ-AI (approved):
### Summary
This PR fixes a real and long-standing crash in `PushConsumer` teardown where the destructor could run on a consume worker thread concurrently with process/static teardown. The fix adds an explicit `shutdown()` API and hardens the internal teardown path against the race conditions.

**Strengths:**
- `PushConsumer::shutdown()` provides a deterministic, owner-thread teardown — the right API design
- `ThreadPoolImpl::shutdown()` self-join detection is correct: detaching the current thread avoids `std::system_error(EDEADLK)` which would `std::terminate` on the `noexcept` path
- Removing `SPDLOG_DEBUG` from `~PushConsumerImpl()` eliminates the use-after-free of the static logger during process teardown
- Null check in `ConsumeTask::process()` after `consumer().lock()` prevents dereferencing an expired consumer
- `shutdown()` is idempotent — safe to call multiple times and from the destructor
- Good regression tests covering the self-join scenario, expired consumer, and idempotent shutdown

**Minor observations:**
- The detached thread in `ThreadPoolImpl::shutdown()` has a clear comment explaining why detach is safe here (worker unwinds once `io_context::run()` returns)
- The 200ms sleep in the test is acceptable for test-only synchronization

LGTM. Important fix for a crash that has been reported multiple times (#421, #521).

lollipopjin (approved):
LGTM

PR #1337, opened 2026-08-13
[ISSUE #1335] [C++] Bound the synchronous producer paths with the request timeout

### LINKED ISSUE #1335: [Bug] [C++] Synchronous producer APIs can block forever; SendMessage has no deadline

lizhimins:

Fixes #1335

### Brief Description

The synchronous producer APIs could block their calling thread indefinitely, and `Configuration::withRequestTimeout()` did not bound them.

**`SendMessage` had no gRPC deadline.** `send()` was the only RPC in the `ClientManager` interface without a `timeout` parameter:

| RPC | `timeout` parameter (before) |
| --- | --- |
| `ack(...)` | yes |
| `changeInvisibleDuration(...)` | yes |
| `endTransaction(...)` | yes |
| `recallMessage(...)` | yes |
| `send(...)` | **no** |

So `ClientManagerImpl::send()` never called `set_deadline()`, `RpcClientImpl::asyncSend()` handed the context to the stub without one, and a stalled broker left the completion callback pending forever. This is a plumbing omission rather than a deliberate design choice, which is why the fix simply gives `send()` the same `timeout` parameter every sibling RPC already has.

**All four synchronous wait sites used `absl::CondVar::Wait()` with no deadline, and three could also lose the wakeup:**

| Site | Completion guard (before) | Deadline (before) |
| --- | --- | --- |
| `ProducerImpl::send()` sync | correct | none |
| `endTransaction0()` | **none at all** | none |
| `recallMessage()` | **no flag even existed** | none |
| `getPublishInfo()` | **racy — read outside the mutex** | none |

`getPublishInfo()` tested `complete` without holding the mutex, so if the callback ran between the check and `Wait()` the signal was dropped and the thread blocked forever with nobody left to wake it — a permanent hang, not a slow call. `endTransaction0()` and `recallMessage()` checked no completion flag at all, so an already-completed callback stranded the caller unconditionally.

Changes:

- Add a `timeout` parameter to `ClientManager::send()` and set the gRPC deadline in `ClientManagerImpl::send()`.
- Pass `requestTimeout()` at the send call site, so `withRequestTimeout()` now bounds normal `SendMessage` RPCs.
- Convert all four synchronous waits to `WaitWithDeadline()` guarded by a completion flag that is only read under the mutex, reporting `RequestTimeout` (or a null publish info) when the deadline expires.

Note on scope: the issue as originally diagnosed covered `send()` and `getPublishInfo()`. While fixing those I found `endTransaction0()` and `recallMessage()` share the same defect in a worse form, so all four are fixed here rather than leaving two known hangs behind.

The interface change is source-compatible for users, since `ClientManager` is internal (`cpp/source/client/include/`) and not part of the public `cpp/include/rocketmq/` surface. The mock and the four affected test call sites are updated accordingly.

Added `ProducerImplTest.sendTimesOutWhenRouteStallsTest`, which stalls the route query (the mock never invokes its callback) and asserts that `send()` returns an error within a bounded time, with a hard 10s cap so a regression fails instead of hanging the suite.

This test was verified to be a real guard: reverting `getPublishInfo()` to the previous unbounded `Wait()` makes the test process hang permanently — it cannot even be reclaimed by its own 10s cap, because the stuck thread can never be joined. That is exactly the reported symptom.

Full C++ suite on this branch, based directly on `master`:

```
$ bazel test //source/...
Executed 36 out of 37 tests: 37 tests pass.
```

This branch contains only this fix and sits directly on `master`, so it can be reviewed and merged independently of my consumer-side fix for #1334.

### DIFF
--- a/cpp/source/client/ClientManagerImpl.cpp
+++ b/cpp/source/client/ClientManagerImpl.cpp
@@ -288,6 +288,7 @@ void ClientManagerImpl::doHeartbeat() {
 bool ClientManagerImpl::send(const std::string& target_host,
                              const Metadata& metadata,
                              SendMessageRequest& request,
+                             std::chrono::milliseconds timeout,
                              SendResultCallback cb) {
   assert(cb);
   SPDLOG_DEBUG("Prepare to send message to {} asynchronously. Request: {}", target_host, request.ShortDebugString());
@@ -296,6 +297,10 @@ bool ClientManagerImpl::send(const std::string& target_host,
   auto invocation_context = new InvocationContext<SendMessageResponse>();
   invocation_context->task_name = fmt::format("Send message to {}", target_host);
   invocation_context->remote_address = target_host;
+  // Bound the SendMessage RPC. Without a deadline a stalled broker leaves the
+  // completion callback pending forever, which in turn blocks the synchronous
+  // send path indefinitely.
+  invocation_context->context.set_deadline(std::chrono::system_clock::now() + timeout);
   for (const auto& entry : metadata) {
     invocation_context->context.AddMetadata(entry.first, entry.second);
   }
--- a/cpp/source/client/include/ClientManager.h
+++ b/cpp/source/client/include/ClientManager.h
@@ -97,6 +97,7 @@ public:
   virtual bool send(const std::string& target_host,
                     const Metadata& metadata,
                     SendMessageRequest& request,
+                    std::chrono::milliseconds timeout,
                     SendResultCallback cb) = 0;
 
   virtual std::error_code notifyClientTermination(const std::string& target_host, const Metadata& metadata,
--- a/cpp/source/client/include/ClientManagerImpl.h
+++ b/cpp/source/client/include/ClientManagerImpl.h
@@ -83,6 +83,7 @@ public:
   bool send(const std::string& target_host,
             const Metadata& metadata,
             SendMessageRequest& request,
+            std::chrono::milliseconds timeout,
             SendResultCallback cb) override LOCKS_EXCLUDED(rpc_clients_mtx_);
 
   /**
--- a/cpp/source/client/mocks/include/ClientManagerMock.h
+++ b/cpp/source/client/mocks/include/ClientManagerMock.h
@@ -89,7 +89,10 @@ public:
                ReceiveMessageCallback),
               (override));
 
-  MOCK_METHOD(bool, send, (const std::string&, const Metadata&, SendMessageRequest&, SendResultCallback), (override));
+  MOCK_METHOD(bool, send,
+              (const std::string&, const Metadata&, SendMessageRequest&, std::chrono::milliseconds,
+               SendResultCallback),
+              (override));
 
   MOCK_METHOD(std::error_code, notifyClientTermination,
               (const std::string&, const Metadata&, const NotifyClientTerminationRequest&, std::chrono::milliseconds),
--- a/cpp/source/client/tests/ClientManagerTest.cpp
+++ b/cpp/source/client/tests/ClientManagerTest.cpp
@@ -16,6 +16,7 @@
  */
 #include <apache/rocketmq/v2/definition.pb.h>
 
+#include <chrono>
 #include <memory>
 #include <system_error>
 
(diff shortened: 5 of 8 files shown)


RockteMQ-AI (approved):
### Summary
This PR correctly fixes the unbounded blocking in synchronous producer paths by adding a `timeout` parameter to `ClientManager::send()` and converting all four synchronous wait sites from `Wait()` to `WaitWithDeadline()`. The fix is well-scoped, source-compatible (internal interface only), and includes a regression test.

**Strengths:**
- Thorough root-cause analysis — identifies not just the reported `send()` issue but also the worse forms in `endTransaction0()`, `recallMessage()`, and the race in `getPublishInfo()`
- The `getPublishInfo()` fix correctly moves the `complete` check inside the mutex, eliminating the signal-drop hang
- Completion flags are consistently read under the mutex across all four sites
- `recall()` adds a `completed` guard to prevent double-invocation of the callback
- Regression test (`sendTimesOutWhenRouteStallsTest`) validates the bounded-timeout behavior with a hard 10s cap

**Minor observations:**
- `getPublishInfo()` timeout message ("Failed to get publish info within timeout") is clear and actionable
- The `recall()` `completed_` flag is correctly initialized to `false` and set under the mutex before invoking the callback

LGTM. Clean, well-tested fix for a class of hang bugs.

lollipopjin (approved):
LGTM

PR #1352, opened 2026-08-28
[Node.js] Fix createDuration producing non-integer protobuf Duration seconds

f1amingo:
### Summary

`createDuration(ms)` sets `Duration.seconds` to `ms / 1000` without flooring.
`Duration.seconds` is int64, so any non-whole-second `ms` (e.g. 2173) makes
serialization throw `Assertion failed` before the RPC leaves the client.

In the SUSPEND path, `ProcessQueue#changeInvisibleDuration` swallows the error
and retries the same invalid duration every second: the RPC never reaches the
broker, the message hangs, and nothing is logged. All `createDuration` call
sites are affected (invisible duration, long-polling and request timeouts).

### Changes

- `createDuration()`: floor seconds; nanos already carry the remainder
- `ProcessQueue`: log the swallowed exception before retrying
- `ConsumeResultSuspend.of()`: reject non-integer ms (also NaN/Infinity)
- Unit tests: `createDuration(2173)` serializes as `(2s, 173000000ns)`;
  suspend accepts 2173, rejects 100.5/NaN

Unit tests in `test/util/index.test.ts` (the original repro
`createDuration(2173).serializeBinary()`) and
`test/consumer/ConsumeResult.test.ts`. Whole-second values are unaffected.

### DIFF
--- a/nodejs/src/consumer/ConsumeResult.ts
+++ b/nodejs/src/consumer/ConsumeResult.ts
@@ -57,6 +57,9 @@ export class ConsumeResultSuspend extends ConsumeResult {
 
   private constructor(suspendTimeMs: number) {
     super('SUSPEND');
+    if (!Number.isInteger(suspendTimeMs)) {
+      throw new Error(`suspend time must be an integer number of milliseconds, got ${suspendTimeMs}`);
+    }
     if (suspendTimeMs < MIN_SUSPEND_TIME_MS) {
       throw new Error(`suspend time cannot be less than ${MIN_SUSPEND_TIME_MS}ms, got ${suspendTimeMs}ms`);
     }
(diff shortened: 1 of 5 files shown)


RockteMQ-AI (approved):
### Summary
This PR fixes a real bug where `createDuration(ms)` could produce non-integer `Duration.seconds` values (e.g., 2173ms → 2.173s), causing protobuf serialization to throw `Assertion failed` before the RPC ever leaves the client. In the SUSPEND path, the error was silently swallowed, causing messages to hang indefinitely with no log output.

The fix is correct and well-scoped:
- **`createDuration`**: Uses `Math.floor()` on both seconds and nanos to ensure protobuf int64 compliance.
- **`ConsumeResultSuspend`**: Adds early integer validation to fail fast with a clear error message.
- **`ProcessQueue`**: Improves error logging in catch blocks — the previous silent `catch {}` was the root cause of the "nothing is logged" symptom described in the issue.
- **Tests**: Good coverage including edge cases (non-whole-second, non-integer, NaN).

Clean fix, LGTM.

---

RongtongJin (approved):

PR #1367, opened 2026-09-08
fix(rust): use mock server address in session_new test

liooooo29:
### Brief Description

Fix the `session::tests::session_new` unit test, which was failing with `Connection refused (os error 111)`.

Root cause: `wiremock-grpc 0.0.3-alpha3` binds its mock gRPC server to the IPv6 loopback address `[::1]`, but the test connected to `localhost:{port}`, which resolves to IPv4 `127.0.0.1` on most Linux hosts. The client therefore dialed IPv4 while the server only listened on IPv6.

Change: construct the connection address from the mock server's actual bound address (`server.address()`) instead of hardcoding `localhost`, so the test is robust regardless of which loopback family the mock binds.

- `cargo test --lib`: 89 passed / 0 failed / 1 ignored (previously 88 passed / 1 failed).
- `cargo fmt --check`: clean.

### DIFF
--- a/rust/src/session.rs
+++ b/rust/src/session.rs
@@ -750,7 +750,7 @@ mod tests {
     async fn session_new() {
         let server = RocketMQMockServer::start_default().await;
         let session = Session::new(
-            &Endpoints::from_url(&format!("localhost:{}", server.address().port())).unwrap(),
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (approved):
LGTM. Trivial change, looks good.

---

PR #1374, opened 2026-09-15
fix(nodejs): fail fast on NotFoundException during startup route fetch

zhaohai666:
## P2 fix(nodejs): fail fast on NotFoundException during startup route fetch

Startup rejects immediately when the initial `QueryRoute` returns `NotFoundException`,
instead of silently retrying forever. Non-not-found errors keep the 3-attempt retry
with linear backoff, so the error message and cause chain now carry
`NotFoundException(40402)` directly.

**Files:** `nodejs/src/client/BaseClient.ts` (1 hunk)


RockteMQ-AI (approved):
### Summary

Fail-fast on `NotFoundException` during startup route fetch — clean, focused fix that aligns with the Java client behavior.

The logic is correct: if the initial `QueryRoute` returns `NotFound`, retrying will never succeed, so throwing immediately is the right call. The added `tsconfig.test.json` is a reasonable addition for test compilation.

CI is green. LGTM.

> **Note:** This PR and #1375/#1376 all add the same `tsconfig.test.json`. If merged sequentially, the later PRs will have a merge conflict on that file. Consider coordinating the merge order or consolidating the file into one PR.

---

RongtongJin (approved):

